PR #1, opened 2024-08-25
Minor shell script updates

ejsears:
- Start to flesh out README.md a bit by adding install / uninstall instructions
- Check to make sure klipper is installed is called
- Added help function
- Add function to show block for moonraker monitoring for updates.


ejsears:
Note, for moonraker, this would require using releases to track changes.

PR #51, opened 2024-10-02
fix(form-tip): copy initial retration settings from happy hare

ammmze:
It looks like the whole tip forming was ported from happy hare but these few lines did not appear to be ported correctly. Original found [here](https://github.com/moggieuk/Happy-Hare/blob/d70ff2096d45b2564cacc0d23139b127cb2c1bc1/config/base/mmu_form_tip.cfg#L99-L101).


MG-longshot:
done

PR #77, opened 2024-10-21
Typo fix in install script

Wondro:


jimmyjon711 (approved):

install-afc.sh
```diff
@@ -106,7 +106,7 @@ check_root() {
 
 restart_service() {
   local service_name=$1
-  print_msh INFO "Restarting ${service_name} service..."
```
ejsears, line 109: urgh, my keyboard is dieing and i need to replace it :(

PR #107, opened 2024-11-10
handle previous lane being current but not loaded during prep

weemantella:
This is an unlikely state but still possible. This will help to cover all possible states during PREP more thoroughly


MG-longshot:
reworking PREP routine

PR #145, opened 2024-11-29
Unit sepration

MG-longshot:


extras/AFC_BoxTurtle.py
```diff
@@ -19,5 +20,74 @@ def __init__(self, config):
         self.logo_error+='R |          |\ <span class=secondary--text>X</span> |\n'
         self.logo_error+='! \_________/ |___|</error>\n'
 
+    def handle_connect(self):
+        """
+        Handle the connection event.
+        This function is called when the printer connects. It looks up the toolhead object
```
jimmyjon711, line 26: Minor thing, comment could be updated to say looks up AFC and assigned to `self.AFC` we can always update this later

jimmyjon711 (approved):

extras/AFC_BoxTurtle.py
MG-longshot: better?

PR #147, opened 2024-11-29
AFC-4 add install versioning

ejsears:
- Adds the ability to check if a full upgrade needs to be done to the config files. This is controlled with the `AFC_VERSION` variable in the `.afc-version` hidden file. This variable is only related to the software upgrade process and must be in SEMVER format (x.y.z).

- Add process to prompt for TN pins (will set to UNKNOWN) if the user doesn't input anything

- Split out some functions for better readability / maintainability

- Prep for fluid/mainsail stuff.


jimmyjon711 (approved):
Pulled down and tested locally

PR #163, opened 2024-12-02
Buffer Ram Sensor

weemantella:
## Major Changes in this PR
- Use of buffer as Ram sensor

## Notes to Code Reviewers
- There is a possible problem with manual lane moves while a lane is loaded this needs to be looked into.

- Follow Buffer_Ram_Sensor doc. load and unload tool.

- [*] I have performed a self-review of my code
- [*] CHANGELOG.md is updated (if end-user facing)
- [*] Sent notification to software-design channel requesting review

### DIFF
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -576,7 +576,6 @@ class afc:
                         return False
 
             # Synchronize lane's extruder stepper and finalize tool loading.
-            CUR_LANE.extruder_stepper.sync_to_extruder(CUR_LANE.extruder_name)
             CUR_LANE.status = 'Tooled'
 
             # Adjust tool position for loading.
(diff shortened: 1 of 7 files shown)


extras/AFC.py
```diff
@@ -584,6 +584,23 @@ def TOOL_LOAD(self, CUR_LANE):
...
+                    CUR_LANE.move( self.short_move_dis * -1, self.short_moves_speed, self.short_moves_accel )
+                    load_checks += 1
+                    self.reactor.pause(self.reactor.monotonic() + 0.1)
+                    if load_checks > self.tool_max_load_checks:
+                        msg = ''
+                        msg += "Buffer did not become compressed after {} short moves.\n".format(self.tool_max_load_checks)
+                        msg += "Tool may not be loaded"
+                        self.gcode.respond_info("<span class=warning--text>{}</span>".format(msg))
+                        break
+                CUR_LANE.extruder_stepper.sync_to_extruder(CUR_LANE.extruder_name)
```
jimmyjon711, line 603: This should probably be outside of the if statement so it syncs back up for either buffer or endstops

PR #168, opened 2024-12-04
Mg longshot

MG-longshot:
 
### DIFF
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -1007,8 +1007,8 @@ class afc:
             for x in range(99):
                 cmd = 'T'+str(x)
                 if cmd not in self.tool_cmds:
-                    self.lanes[CUR_LANE.unit][CUR_LANE.name]['map'] = 'T'+str(x)
-                    CUR_LANE.map = 'T'+str(x)
+                    self.lanes[CUR_LANE.unit][CUR_LANE.name]['map'] = cmd
+                    CUR_LANE.map = cmd
                     break
         self.tool_cmds[self.lanes[CUR_LANE.unit][CUR_LANE.name]['map']]=CUR_LANE.name
         try:
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -1002,5 +1002,21 @@ class afc:
             self.gcode.respond_info( error_string)
         led.led_change(int(idx.split(':')[1]), status)
 
+    def TcmdAssign(self, CUR_LANE):
+        if CUR_LANE.map == 'NONE' :
+            for x in range(99):
+                cmd = 'T'+str(x)
+                if cmd not in self.tool_cmds:
+                    self.lanes[CUR_LANE.unit][CUR_LANE.name]['map'] = 'T'+str(x)
+                    CUR_LANE.map = 'T'+str(x)
+                    self.gcode.respond_info(CUR_LANE.name + ' Mapped to ' + 'T'+str(x))
+                    break
+        self.tool_cmds[self.lanes[CUR_LANE.unit][CUR_LANE.name]['map']]=CUR_LANE.name
+        try:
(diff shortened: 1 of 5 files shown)


extras/AFC_prep.py
```diff
@@ -94,7 +95,13 @@ def PREP(self, gcmd):
                 if 'color' not in self.AFC.lanes[LANE.unit][LANE.name]: self.AFC.lanes[LANE.unit][LANE.name]['color']='#000000'
                 if 'weight' not in self.AFC.lanes[LANE.unit][LANE.name]: self.AFC.lanes[LANE.unit][LANE.name]['weight'] = 0
                 if 'runout_lane' not in self.AFC.lanes[LANE.unit][LANE.name]: self.AFC.lanes[LANE.unit][LANE.name]['runout_lane']='NONE'
-                if 'map' not in self.AFC.lanes[LANE.unit][LANE.name] or self.AFC.lanes[LANE.unit][LANE.name]['map'] is None: self.AFC.lanes[LANE.unit][LANE.name]['map'] = LANE.map
+                if 'map' not in self.AFC.lanes[LANE.unit][LANE.name]:
```
jimmyjon711, line 98: Should probably keep `or self.AFC.lanes[LANE.unit][LANE.name]['map'] is None:` check in just in case someones config file gets map set to null. This will cover the null cases
```suggestion
                if 'map' not in self.AFC.lanes[LANE.unit][LANE.name] or self.AFC.lanes[LANE.unit][LANE.name]['map'] is None::
```

MG-longshot: the config input in the stepper makes it 'NONE' if null

extras/AFC_prep.py
```diff
@@ -94,7 +95,13 @@ def PREP(self, gcmd):
...
+                   self.AFC.lanes[LANE.unit][LANE.name]['map'] = 'NONE'
+                else:
+                   LANE.map = self.AFC.lanes[LANE.unit][LANE.name]['map']
+                if LANE.map != 'NONE':
```
jimmyjon711, line 102: This if statement is not needed same thing with line 103 since its just reassigning the value back to the array that was assigned to `LANE.map` in the else statement above, just need to keep line 104 within the else statement

extras/AFC.py
```diff
@@ -1002,5 +1002,20 @@ def afc_led (self, status, idx=None):
             self.gcode.respond_info( error_string)
         led.led_change(int(idx.split(':')[1]), status)
 
+    def TcmdAssign(self, CUR_LANE):
+        if CUR_LANE.map == 'NONE' :
+            for x in range(99):
+                cmd = 'T'+str(x)
+                if cmd not in self.tool_cmds:
+                    self.lanes[CUR_LANE.unit][CUR_LANE.name]['map'] = 'T'+str(x)
```
jimmyjon711, line 1010: Use the cmd variable that was created on line 1008
```suggestion
                    self.lanes[CUR_LANE.unit][CUR_LANE.name]['map'] = cmd
```

extras/AFC.py
```diff
@@ -1002,5 +1002,20 @@ def afc_led (self, status, idx=None):
...
+                    CUR_LANE.map = 'T'+str(x)
```
jimmyjon711, line 1011: Use the cmd variable that was created on line 1008
```suggestion
                    CUR_LANE.map = cmd
```

extras/AFC.py
MG-longshot: Done

extras/AFC.py
MG-longshot: again done

extras/AFC_prep.py
jimmyjon711: But if for some weird reason this `"map": null` gets stuck in someones AFC.var.unit file it will error out. I agree it should not happen but its always the cases we don't account for that breaks things

MG-longshot: added back in

jimmyjon711 (approved):

PR #181, opened 2024-12-10
Mg longshot

MG-longshot:
 
### DIFF
--- a/extras/AFC_BoxTurtle.py
+++ b/extras/AFC_BoxTurtle.py
@@ -94,5 +94,321 @@ class afcBoxTurtle:
 
         return succeeded
 
+    def TOOL_LOAD(self, CUR_LANE):
+        """
+        This function handles the loading of a specified lane into the tool. It performs
+        several checks and movements to ensure the lane is properly loaded.
+
+        Usage: `TOOL_LOAD LANE=<lane>`
+        Example: `TOOL_LOAD LANE=leg1`
+
+        Args:
+            CUR_LANE: The lane object to be loaded into the tool.
+
+        Returns:
+            bool: True if load was successful, False if an error occurred.
+        """
+        if CUR_LANE is None:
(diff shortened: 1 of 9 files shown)


extras/AFC_BoxTurtle.py
```diff
@@ -94,5 +96,482 @@ def system_Test(self, UNIT, LANE, delay):
...
+        """
+        if CUR_LANE is None:
+            # Exit early if no lane is provided.
+            return False
+
+        # Check if the bypass filament sensor is triggered; abort loading if filament is already present.
+        try:
+            bypass = self.printer.lookup_object('filament_switch_sensor bypass').runout_helper
+            if bypass.filament_present:
+                self.gcode.respond_info("Filament loaded in bypass, not doing tool load")
```
jimmyjon711, line 286: Nees to be updated to be `self.AFC.gcode`

extras/AFC_hub.py
```diff
@@ -47,10 +47,76 @@ def handle_connect(self):
         self.gcode = self.AFC.gcode
         self.reactor = self.AFC.reactor
 
+        #self.gcode.register_command('HUB_CUT_TEST', self.cmd_HUB_CUT_TEST, desc=self.cmd_HUB_CUT_TEST_help)
```
jimmyjon711, line 50: Uncomment both lines?

jimmyjon711 (changes requested):

extras/AFC_hub.py
jimmyjon711: So thinking about this more, wouldn't it make sense to keep the register commands in AFC.py and a cmd_ function in AFC.py that then looks up current lane and calls CUR_LANE.unit.hub_cut_test / hub_load functions. This would be the same way that TOOL_LOAD/UNLOAD would work

PR #220, opened 2025-01-09
Fix hub cutting so grip on filament is not lost; add assisted retracts to hub cutting

rschaeuble:
## Major Changes in this PR

The long retract (4*move_dis) in the hub cut sequence moves the filament to the entry of the extruder. This has a large risk of losing grip on the filament (happend in roughly 30 % of attempts for me).
Using the hub as reference point instead of extruder entry fixes that completely.

However, this requires users to update their configuration! Please consider announcing this in the appropriate Discord channel, so noone is surprised.

I also added assisted retracts as an option during hub cuts, so the filament stays tight on the spol

I performed my serial number print with this modification. Worked perfectly, which before I again and again had to intervene when grip was lost.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -301,3 +301,17 @@ file as the code now generates them automatically.
   Allows to update tip forming configuration at runtime.
 
   See command_reference doc for more info
+
+## [2025-01-10]
+
+### Changed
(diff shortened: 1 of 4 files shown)


rschaeuble:
So what is this software-design channel? I can't see such a channel in Discord.

jimmyjon711 (approved):

PR #247, opened 2025-01-28
Adding Purge Length Support

ImSundee:
## Major Changes in this PR
This pull adds support for PURGE_LENGTH to a change. 

If PURGE_LENGTH is not supplied it will not parse one, so will use the default amount.

in Orca your change filament code to support this would now be:
_T[next_extruder] PURGE_LENGTH=[flush_length]_

## Notes to Code Reviewers

Sundee on discord for any questions

I have been running this on my machine with over 1000+ filament changes now.

I have not updated change log so you can update it if this is merged.


ImSundee:
Removing as missed half the commit

PR #254, opened 2025-02-13
New feature: assisted unload

rschaeuble:
## Major Changes in this PR

New feature: assisted unload

When enabled, the retracts out of the toolhead before the long, fast move back throught the bowden tube is assisted. This helps with full spools where even a retract of a few centimeters can cause a loop to fall off the spool.

**TBD**: do we want to enable this by default? I don't really see a reason why not. In the worst case it neither helps nor harms, in the best case users don't have to find out that they need this themselves.

By setting the flag on AFC, unit and extruder level and checking if on unload the spool is turned by the respoolers.

### DIFF
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -129,15 +129,17 @@ class afc:
         self.tool_max_unload_attempts = config.getint('tool_max_unload_attempts', 2)# Max number of attempts to unload filament from toolhead when using buffer as ramming sensor
         self.tool_max_load_checks = config.getint('tool_max_load_checks', 4)        # Max number of attempts to check to make sure filament is loaded into toolhead extruder when using buffer as ramming sensor
 
-        self.z_hop =config.getfloat("z_hop", 0)                                     # Height to move up before and after a tool change completes
-        self.xy_resume =config.getboolean("xy_resume", False)                       # Need description or remove as this is currently an unused variable
-        self.resume_speed =config.getfloat("resume_speed", 0)                       # Speed mm/s of resume move. Set to 0 to use gcode speed
+        self.z_hop = config.getfloat("z_hop", 0)                                     # Height to move up before and after a tool change completes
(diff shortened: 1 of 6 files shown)


extras/AFC.py
```diff
@@ -129,15 +129,17 @@ def __init__(self, config):
...
 
         self.global_print_current = config.getfloat("global_print_current", None)   # Global variable to set steppers current to a specified current when printing. Going lower than 0.6 may result in TurtleNeck buffer's not working correctly
 
         self.enable_sensors_in_gui = config.getboolean("enable_sensors_in_gui", False) # Set to True to show all sensor switches as filament sensors in mainsail/fluidd gui
-        self.load_to_hub        = config.getboolean("load_to_hub", True)            # Fast loads filament to hub when inserted, set to False to disable. This is a global setting and can be overridden at AFC_stepper
+        self.load_to_hub = config.getboolean("load_to_hub", True)            # Fast loads filament to hub when inserted, set to False to disable. This is a global setting and can be overridden at AFC_stepper
+        self.assisted_unload = config.getboolean("assisted_unload", False)   # If True, the unload retract is assisted to prevent loose windings, especially on full spools. This can prevent loops from slipping off the spool. Setting value here overrides values set in AFC.cfg file
```
jimmyjon711, line 141: Can you please manually add this variable and comment to `docs/CONFIGURATION_OPTIONS.md` document for AFC, AFC_BoxTurtle and AFC_NightOwl

rschaeuble: I added the options.

rschaeuble:
So, should we enable this feature by default or not?

Pro:
- It prevents loops from falling of a spool, so it makes the print more reliable.
- Most users will probably find out about this option *after* they had a problem. Enabling it by default will save them from having this problem in the first place.

Con:
- Might theoretically break something for someone, as always when behavior changes due to an update.

jimmyjon711 (approved):

jimmyjon711 (approved):

PR #289, opened 2025-03-26
Fix Effective Diameter Calculation

jojoa:
## Major Changes in this PR
The Effective Diameter Calculation was wrong implemented, so i corrected it.
## Notes to Code Reviewers
The Factor of 0.785 is theoretical packages density.
Tested the calculation with 2 of my spools by hand.
 
### DIFF
--- a/extras/AFC_stepper.py
+++ b/extras/AFC_stepper.py
@@ -653,10 +653,9 @@ class AFCExtruderStepper:
 
         # Calculate the cross-sectional area of the filament
         density_g_mm3 = self.filament_density / 1000.0
-        filament_cross_section_mm2 = 3.14159 * (self.filament_diameter / 2) ** 2
         filament_volume_mm3 = weight_g / density_g_mm3
-        filament_length_mm = filament_volume_mm3 / filament_cross_section_mm2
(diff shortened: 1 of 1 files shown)


jimmyjon711:
Please change branch to push into DEV not main

jojoa:
Rebased to DEV Head

jimmyjon711:
What was the problem originally that made you look into doing this change, and what does this change fix? Also where did you get 0.785 number from.

jojoa:
The Problem was that the rotational speed for the rewinder/assist was wrong on a nearly empty spool. So i looked into the code, and found this wrong calculation. The Change fixes the calculation for die outer diameter of the filament on the spool. the Factor 0.785 is the factor for the cubical dense cylindrical packackage is comes from pi/4, and is needed as we stack the filament(cylinder) onto eachother.

weemantella:
Can you share more data about how the performance has changed/improved?

jojoa:
I am not sure that to show out in the code. And for the performance i tested 2 of my roles that were off by several centimeters and now i am within 1 cm in diameter. Which means the assist roller are working better and are less slipping due to wrong speed.
picture
Here is a pic of the packaging. The square is a^2 and the brown area which represent the filament is (d/2)^2*pi so the factor is pi/4

PR #326, opened 2025-04-27
Refactor string concatenation to use .format()

Surion79:
## Major Changes in this PR
Just refactored string concat to be unified to .format() usage

changes are based only on strings from + concat to format as it is used in the rest of the project

- [X] CHANGELOG.md is updated (if end-user facing)
- [X] Documentation updated in AT-Documentation repository
- [] Sent notification to software-design channel requesting review

### DIFF
--- a/extras/AFC_functions.py
+++ b/extras/AFC_functions.py
@@ -108,7 +108,7 @@ class afcFunction:
     def TcmdAssign(self, CUR_LANE):
         if CUR_LANE.map == 'NONE' :
             for x in range(99):
-                cmd = 'T'+str(x)
+                cmd = 'T{0}'.format(x)
                 if cmd not in self.AFC.tool_cmds:
                     CUR_LANE.map = cmd
                     break
@@ -215,7 +215,7 @@ class afcFunction:
         """
         error_string = ""
         led = None
-        afc_object = 'AFC_led '+ led_name.split(':')[0]
+        afc_object = 'AFC_led {}'.format(led_name.split(':')[0])
         try:
             led = self.printer.lookup_object(afc_object)
         except:
@@ -236,10 +236,10 @@ class afcFunction:
         if CUR_LANE.prep_state:
             if CUR_LANE.load_state:
                 if CUR_LANE.extruder_obj is not None and CUR_LANE.extruder_obj.lane_loaded == CUR_LANE.name:
-                    return 'In Tool:' + self.HexConvert(CUR_LANE.led_tool_loaded).split(':')[-1]
-                return "Ready:" + self.HexConvert(CUR_LANE.led_ready).split(':')[-1]
-            return 'Prep:' + self.HexConvert(CUR_LANE.led_prep_loaded).split(':')[-1]
-        return 'Not Ready:' + self.HexConvert(CUR_LANE.led_not_ready).split(':')[-1]
+                    return 'In Tool:{}'.format(self.HexConvert(CUR_LANE.led_tool_loaded).split(':')[-1])
+                return "Ready:{}".format(self.HexConvert(CUR_LANE.led_ready).split(':')[-1])
+            return 'Prep:{}'.format(self.HexConvert(CUR_LANE.led_prep_loaded).split(':')[-1])
+        return 'Not Ready:{}'.format(self.HexConvert(CUR_LANE.led_not_ready).split(':')[-1])
 
     def handle_activate_extruder(self):
         """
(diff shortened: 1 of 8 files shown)


extras/AFC_functions.py
```diff
@@ -399,7 +399,7 @@ def cmd_ALL_CALIBRATION(self, gcmd):
         prompt = AFCprompt(gcmd, self.logger)
         footer = []
         title = 'Calibrate all'
-        text = ('Press Yes to confirm calibrating all lanes in all units')
```
jimmyjon711, line 402: @weemantella Are the parentheses still needed to show the calibration messages correctly?

extras/AFC_prep.py
```diff
@@ -69,8 +69,8 @@ def PREP(self, gcmd):
 
         ## load Unit stored variables
         units={}
-        if os.path.exists(self.AFC.VarFile + '.unit') and os.stat(self.AFC.VarFile + '.unit').st_size > 0:
-            units=json.load(open(self.AFC.VarFile + '.unit'))
+        if os.path.exists('{}.unit'.format(self.AFC.VarFile)) and os.stat('{}.unit'.format(self.AFC.VarFile)).st_size > 0:
+            units=json.load(open('{}.unit'.format(self.AFC.VarFile)))
```
jimmyjon711, line 73: Since this is being changed, I think it makes more sense to make this filename a variable and use the variable in these other checks?

jimmyjon711 (changes requested):

extras/AFC_functions.py
weemantella: No it should be fine without parentheses I believe. As long as it's tested after it's changed that will confirm.

extras/AFC_functions.py
```diff
@@ -611,7 +611,7 @@ def cmd_AFC_HAPPY_P(self, gcmd):
         buttons = None
         footer = []
         title = '{} Completed'.format(step)
-        text = ('Happy Printing!')
+        text = 'Happy Printing!'
         footer.append(('EXIT', 'prompt_end', 'info'))
```
weemantella, line 615: Can you remove the footer button? Also just remove footer from the function? It causes potential confusion.

extras/AFC_functions.py
jimmyjon711: @Surion79 can you run through just one lane calibration just to verify it will work properly still.

extras/AFC_prep.py
Surion79: should i make it a class attribute?

extras/AFC_functions.py
Surion79: done

Surion79:
awaiting feedback on the var file and the other PR what to do before doing the requested testing

extras/AFC_prep.py
Surion79: or other question: shouldn't the variable VarFile include ".unit" that would be the most efficient/understandable one

Surion79: or even that the AFC.cfg is being updated to have the correct value from the beginning? ;) 
That could be handled to check if the value has already ending with .unit (for new config then) with a default to add .unit if it is an old config which does not include.unit

Surion79: My second part of my last message took that issue in consideration. Shouldn't be an issue when the automated install script is live anyway?

jimmyjon711: At this point we would not be able to rely on it being in the config file without users having to update their configs. It's this way because two files used to be created based off the varfile path/name and a different extension was added to the end. I think the best thing to do would be to add the `.unit` in the AFC class and then we could drop the `.unit` everywhere and just reference the varfile variable

jimmyjon711: > or even that the AFC.cfg is being updated to have the correct value from the beginning? ;) That could be handled to check if the value has already ending with .unit (for new config then) with a default to add .unit if it is an old config which does not include.unit

Just read this after my first comment, this could work as well I think.

extras/AFC_functions.py
Surion79: i tested a all calibration and did not find any malformation of text

extras/AFC_prep.py
Surion79: i'd like to do that in a dedicated PR, so we can keep this. it works (tested multiple prints with it)

jimmyjon711: I am fine with that, can be done after #330 is merged as well.

jimmyjon711 (approved):
Looks good to me now.

PR #330, opened 2025-04-27
AFC-62 readable error for varfile

ejsears:
## Major Changes in this PR

This pull request makes a minor update to the `PREP` method in `extras/AFC_prep.py` to improve error handling and fix some typos in comments.

Error handling improvements:
* Added a check for the existence of the `.unit` file. If the file is missing, an error message is generated and logged using the `AFC_error` method.

Comment corrections:
* Fixed typos in comments, such as changing "suppose" to "supposed" and correcting "its its" to "it's".

## Notes to Code Reviewers

Do we need changelog entries for minor bugfixes like this?

Make a bad path in the `AFC.cfg` file for the `VarFile` and start up Klipper. It should fail and also `return` out of the function. Otherwise, the `PREP` function would continue, which would result in an internal error. 

image

Fix the path, and it works like normal

image

### DIFF
--- a/extras/AFC_prep.py
+++ b/extras/AFC_prep.py
@@ -71,13 +71,17 @@ class afcPrep:
         units={}
         if os.path.exists(self.AFC.VarFile + '.unit') and os.stat(self.AFC.VarFile + '.unit').st_size > 0:
             units=json.load(open(self.AFC.VarFile + '.unit'))
+        else:
+            error_string = 'Error: ' + self.AFC.VarFile + '.unit file not found. Please check the path in the'
+            error_string += 'AFC.cfg file and make sure the file exists.'
+            self.AFC.ERROR.AFC_error(error_string, False)
 
-        # check if Lane is suppose to be loaded in tool head from saved file
+        # check if Lane is supposed to be loaded in tool head from saved file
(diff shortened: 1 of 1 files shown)


extras/AFC_prep.py
```diff
@@ -71,13 +71,17 @@ def PREP(self, gcmd):
         units={}
         if os.path.exists(self.AFC.VarFile + '.unit') and os.stat(self.AFC.VarFile + '.unit').st_size > 0:
             units=json.load(open(self.AFC.VarFile + '.unit'))
+        else:
+            error_string = 'Error: ' + self.AFC.VarFile + '.unit file not found. Please check the path in the'
```
jimmyjon711, line 75: Please update to use `format()` or f string, trying to remove all the concatenations. Also see my comment on this PR about the varfile https://github.com/ArmoredTurtle/AFC-Klipper-Add-On/pull/326#discussion_r2062732523.

jimmyjon711 (changes requested):

extras/AFC_prep.py
ejsears: Will hold off until that is merged.

jimmyjon711 (approved):

PR #339, opened 2025-05-06
Add support for led range per lane

N3MI-DG:
## Major Changes in this PR
Gives the user the option to use multi pixel led boards for each lane while keeping the existing behavior in tact.
PXL_20250505_064542275 MP

Add a `-` between the first and last pixel number. For example: ```led_index: AFC_Indicator:1-21```
 
### DIFF
--- a/extras/AFC_led.py
+++ b/extras/AFC_led.py
@@ -128,7 +128,12 @@ class AFCled:
             else:
                 set_color_fn = self.led_helper.set_color
                 check_transmit_fn = self.led_helper.check_transmit
-            set_color_fn(index, colors)
+            if isinstance(index, str) and "-" in index:
+                start, end = map(int, index.split("-"))
+                for i in range(start, end + 1):
+                    set_color_fn(i, colors)
+            else:
+                set_color_fn(int(index), colors)
             if transmit:
                 check_transmit_fn(print_time)
         toolhead = self.printer.lookup_object('toolhead')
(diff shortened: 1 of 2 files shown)


extras/AFC_led.py
```diff
@@ -128,21 +128,31 @@ def lookahead_bgfunc(print_time):
...
+
             if transmit:
                 check_transmit_fn(print_time)
         toolhead = self.printer.lookup_object('toolhead')
         toolhead.register_lookahead_callback(lookahead_bgfunc)
 
     def turn_off_leds(self):
-        for i in range(self.led_helper.led_count):
-            self.led_change( i, "0,0,0,0", False)
+        for lane in self.AFC.lanes:
```
jimmyjon711, line 145: Recently these variables were updated to be lowercase, please update. Also can you please test to make sure `TURN_OFF_AFC_LED` and `TURN_ON_AFC_LED` macros work correctly
```suggestion
        for lane in self.afc.lanes:
```

extras/AFC_led.py
```diff
@@ -128,21 +128,31 @@ def lookahead_bgfunc(print_time):
             else:
                 set_color_fn = self.led_helper.set_color
                 check_transmit_fn = self.led_helper.check_transmit
-            set_color_fn(index, colors)
+
+            if "-" in index:
```
jimmyjon711, line 132: This should be the right change to make and revert changes in `turn_off_leds` and `turn_on_leds` functions
```suggestion
            if isinstance(index, str) and "-" in index:
```

jimmyjon711 (changes requested):

N3MI-DG:
Updated the commit with the suggested changes. Thanks for your input @jimmyjon711

jimmyjon711 (approved):
Looks good, tested with my setup as well. Looks like everything works, thanks for making these updates.

PR #399, opened 2025-05-11
added support for per lane rewind speed factor, and variable extruder…

rsarwar87:
added two extra config in the AFC.cfg and AFC_device.cfg
```
tool_homing_distance: 200       # distance over which toolhead homing is done. Default is 200mm.
rev_long_moves_speed_factor: 1.0   # scalar factor for reverse long moves (range - 0.5 - 1.2)
```

```rev_long_moves_speed_factor``` can be overridden from  AFC_stepper
```tool_homing_distance``` can be useful if the filament stalls.

helps with filamentalist based rewinder, where one may need to slow down the rewind speed to allow stalling.

on my quantrobox.


jimmyjon711:
@rsarwar87 Please change target branch from main to DEV

jimmyjon711:
@rsarwar87 actually it might be better to close this PR and checkout DEV and add these changes directly to your DEV branch so you don't merge main into DEV with this PR

PR #422, opened 2025-05-30
Add park macro option to lower z height

antoinecellerier:
This is intended to reduce oozing during filament unloading and loading prior to using the poop macro.

Adds a park_z macro variable to set the z height to move to after the move to the park x,y location.
This is default disabled and can only be used to move closer to the bed.


jimmyjon711 (approved):
Looks good, and testing fine.

PR #463, opened 2025-06-22
Material temp case sensitivity fix

### LINKED ISSUE #423: [Bug]: Case insensitive check when setting temperature
### What happened?

Currently if someone puts all uppercase (eg PETG) as material name in `default_material_temps` list but sets their lanes material as lowercase (eg petg), their extruder temperature will not be set correctly to the temp that they specified in the `default_material_temps` variable.

### Version

1.0.11

### Relevant log output

```shell

```

atgnats:
## Major Changes in this PR

- Fixes #423 : Ensure material name matching in temperature selection logic is case-insensitive and ignores excess whitespace.

## Notes to Code Reviewers

- No functional changes beyond the bugfix; behavior should match previous releases except for improved reliability with case/whitespace.

- Manual test: Verified that the correct default temperature is selected for a lane with material name variations (e.g., "abs", "ABS", " Abs ").
- Confirmed that the updated file is loaded and used by Klipper after a reboot.


jimmyjon711 (approved):
Changes look good, thanks for fixing this.

PR #466, opened 2025-06-22
Fix crashes into depressor arm for filament cutter with fwd/rwd motion

newphreak:
The current macro worked for left to right cutting action. But for forwards/backwards motion it resulted in a crash as it got on the wrong side of the depressor pin. This fixes that.


jimmyjon711:
Please change merge target to DEV and update changelog

PR #475, opened 2025-06-29
Custom load sequence and loaded but idle led status

weemantella:
## Major Changes in this PR
This PR introduces the ability to set a custom load/unload command per lane. The primary use is for systems that function outside of the typical AFC configuration. This allows for a custom command to be called that will override the standard load sequence.
    - If defined this will override and bypass the standard load/unload sequence
    - A pre-extruder filament sensor is currently required
    - This is defined under the `[AFC_lane]` or `[AFC_stepper]` as `custom_load_cmd`/`custom_unload_cmd`
- The second feature that is added is a LED status for when a tool is loaded but is not the active extruder. This is for use in toolchangers for a more clear led status for the system
    - This can be set globally, at the unit, or lane level.
    - `led_tool_loaded_idle`

## Notes to Code Reviewers
This is pretty straightforward. The standard flow is not interrupted if the command is not defined.

1. Define a custom load command and execute it
2. Run a standard load

- Both are successful, and the functionality holds

### DIFF
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -1027,7 +1027,6 @@ class afc:
         if self.function.get_current_extruder() != cur_lane.extruder_obj.name:
             self.tool_swap(cur_lane)
 
-
         if cur_lane.name != self.current:
             # Lookup extruder and hub objects associated with the lane.
             cur_hub = cur_lane.hub_obj
@@ -1119,6 +1118,8 @@ class afc:
         during the loading process.
 
         :param cur_lane: The lane object to be loaded.
+        :param cur_hub: The hub object associated with the lane.
+        :param cur_extruder: The extruder object associated with the lane.
         """
         # Placeholder for custom load sequence
         if cur_lane.custom_load_cmd:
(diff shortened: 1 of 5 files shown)


extras/AFC_lane.py
```diff
@@ -139,6 +143,9 @@ def __init__(self, config):
             buttons.register_buttons([self.load], self.load_callback)
         else: self.load_state = True
 
+        # Custom Load Command
```
jimmyjon711, line 146: Either this PR or a different one can we move this command get up with the others so they are all grouped together

extras/AFC.py
```diff
@@ -1207,6 +1113,132 @@ def TOOL_LOAD(self, cur_lane, purge_length=None):
...
+            self.logger.info("Running custom load command for lane {}".format(cur_lane.name))
+            self.gcode.run_script_from_command(cur_lane.custom_load_cmd)
+            if cur_lane.get_toolhead_pre_sensor_state():
+                cur_lane.status = AFCLaneState.TOOL_LOADED
+                self.save_vars()
+            else:
+                message = 'Custom load command did not trigger pre extruder gear toolhead sensor, CHECK FILAMENT PATH\n||=====||====||==>--||\nTRG   LOAD   HUB   TOOL'
+                message += '\nTo resolve set lane loaded with `SET_LANE_LOADED LANE={}` macro.'.format(cur_lane.name)
+                message += '\nManually move filament until filament is right before toolhead extruder gears,'
+                message += '\n then load into extruder gears with extrude button in your gui of choice until the color fully changes'
```
jimmyjon711, line 1134: Remove space
```suggestion
                message += '\nthen load into extruder gears with extrude button in your gui of choice until the color fully changes'
```

jimmyjon711 (changes requested):
Couple of change to make 🙂

jimmyjon711 (approved):
LGTM

PR #490, opened 2025-07-20
Add SET_NEXT_SPOOL_ID command

kekiefer:
## Major Changes in this PR

Adds a `SET_NEXT_SPOOL_ID` command that can be used by a QR scanning utility to prep the AFC for the spool_id in the next load. This allows a user to load spools entirely at the machine without a separate manual step of keeping track of loaded spool numbers and entering them in mainsail.

## Notes to Code Reviewers

See here for the service that I'm using to automate the sending this command when QR codes are scanned: https://github.com/kekiefer/afc-spool-scan

Manually run the SET_NEXT_SPOOL_ID to any ID registered in spoolman. One spool can be prepped at a time, and the next load to any lane should be automatically set up with that ID. This manual workflow will mirror what is automated through the separate service.

### DIFF
--- a/extras/AFC_spool.py
+++ b/extras/AFC_spool.py
@@ -10,6 +10,9 @@ class AFCSpool:
         self.printer = config.get_printer()
         self.printer.register_event_handler("klippy:connect", self.handle_connect)
 
+        # Temporary status variables
+        self.next_spool_id      = ''
+
     def handle_connect(self):
         """
         Handle the connection event.
@@ -26,6 +29,7 @@ class AFCSpool:
         self.printer.register_event_handler("afc_stepper:register_macros",self.register_lane_macros)
 
         self.gcode.register_command("RESET_AFC_MAPPING", self.cmd_RESET_AFC_MAPPING, desc=self.cmd_RESET_AFC_MAPPING_help)
+        self.gcode.register_command("SET_NEXT_SPOOL_ID", self.cmd_SET_NEXT_SPOOL_ID, desc=self.cmd_SET_NEXT_SPOOL_ID_help)
 
     def register_lane_macros(self, lane_obj):
         """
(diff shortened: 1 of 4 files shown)


kekiefer:
Thanks for catching that, fixed

extras/AFC_spool.py
```diff
@@ -247,6 +251,19 @@ def _get_filament_values( self, filament, field, default=None):
             value = filament[field]
         return value
 
+    def _set_values(self, cur_lane):
+        """
+        Helper function for setting lane spool values
+        """
+        # set defaults if there's no spool id, or the spoolman lookup fails
+        self.material = self.afc.default_material_type
```
jimmyjon711, line 259: The following two lines should be `cur_lane.material` and `cur_lane.weight` so that default material and weight gets set correctly for those that don't use spoolman.

ejsears (approved):
I'm overall good with this. I need to take a look at the service / systemd part of it.

@kekiefer do you want that part merged into this repo also? I could probably add something to the installer to auto install it if it is selected.

ejsears:
@kekiefer Take the copilot stuff with a grain of salt.

kekiefer:
> @kekiefer do you want that part merged into this repo also? I could probably add something to the installer to auto install it if it is selected.

I think it would be nice, as it eliminates extra steps that users would need to take to make use of this feature that could be automated. It will make the installation logic more complicated as there's a couple permissions steps to give the user access to the evdev devices (adding the user to a group), and evtest has to be installed. /dev/hidraw* was the other possibility here, but debian distros don't give anyone but root access to those.

extras/AFC_spool.py
kekiefer: Doh! I just read this again and see what I did wrong here. Yeah I'll fix this up.

jimmyjon711:
@kekiefer Can you also please work on a PR to update the https://github.com/ArmoredTurtle/AT-Documentation repo with the new macro an add if you would like add something to the feature section.

kekiefer:
> @kekiefer Can you also please work on a PR to update the https://github.com/ArmoredTurtle/AT-Documentation repo with the new macro and if you would like add something to the feature section.

Yes, I'll do that.

Thanks guys!

Issue #331, opened 2025-05-01
Buffer only setup: failed loads trigger a warning and not an error (resulting in a failed print)


Surion79:
When having a "buffer only" setup without toolhead sensors, failed loads result in a warning and not in an error to fix manual.
With warnings, the print continues without filament, since there is no error where it is possible to fix the failed load.

PR #492, opened 2025-07-20
bypass tool max unload attempts

kekiefer:
## Major Changes in this PR

#331 caused a regression for toolheads with a cutter above the extruder gears and no toolhead filament sensor, where the buffer doesn't get compressed when the filament is unloaded. This change adds the ability to set tool_max_unload_attempts to zero for this special case to bypass these checks.


jimmyjon711 (approved):
Looks good to me, testing without problems.

Issue #530, opened 2025-09-04
[Enhancement]: Support for cutters placed before extruder wheels


lukaszdrukasz:
### What are you looking to see?

The current implementation (as I understand it) assumes that the filament cutter - if present - is located "after" the extruder wheels. The unload procedure performs the cut and then retracts tool_stn_unload mms of the filament. In my printer (Qidi Plus4) the cutter is located _before_ extruder wheels. The retracted filament fails to clear the filament sensor and is reported to be stuck in the nozzle.

An option I would like to see supported is - for cutter-before-extruder printers - to disengage the main extruder and make only a lane move by negative tool_stn_unload mms instead. The filament could then actually move out of the sensor area and wouldn't be reported as stuck.

I made it work by making cmd_LANE_MOVE not to care about whether a print is active and by inserting a negative LANE_MOVE into the cut macro. 

### Why do you want this feature?

Allows for support of a wider variety of printers

PR #541, opened 2025-10-01
Handle tool_stn_unload as zero for printers with cutter above extruder gears

kekiefer:
## Major Changes in this PR

afc: tool_unload: handle special case of tool_stn_unload == 0

This change allows tool_stn_unload to be zero, to enter a special
unload mode for toolheads with a cutter above the extuder gears and
a filament sensor above both of them.

Resolves https://github.com/ArmoredTurtle/AFC-Klipper-Add-On/issues/530

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2025-09-30]
+### Added
+- Allow `tool_stn_unload` to be `0` for toolheads with cutter above extruder.
+
 ## [2025-09-26]
 ### Added
 - Support to move filament to TD-1 device that is inline with PTFE tube to gather TD and color


extras/AFC.py
```diff
@@ -1377,11 +1377,26 @@ def TOOL_UNLOAD(self, cur_lane):
...
+                cur_lane.unsync_to_extruder()
+                while cur_lane.get_toolhead_pre_sensor_state():
+                    # attempt to move filament back from sensor without moving extruder
+                    cur_lane.move_advanced(cur_lane.short_move_dis * -1, SpeedMode.SHORT)
+                    num_tries += 1
+                    if num_tries > self.tool_max_unload_attempts:
+                        # note that this will break out of the loop and immediately fall into the error
+                        # condition of the next loop for messaging to the user
+                        break
+                    self.reactor.pause(self.reactor.monotonic() + 0.1)
```
jimmyjon711, line 1398: Is this pause needed? Since when calling `cur_lane.move_advanced` the move will wait for all moves to finish before moving on and checking the status of the sensor again.

kekiefer: I did it this way following the existing pattern for the buffer move, but considering what you say, I'm not entirely sure why it was done this way either:
https://github.com/ArmoredTurtle/AFC-Klipper-Add-On/blob/6c8a75daa6ca74a4dfac948fb3781347859a4003/extras/AFC.py#L1361-L1363
Should we use a different pattern here?

PR #580, opened 2025-12-03
Fix lane_move current issue

### LINKED ISSUE #578: [Bug]: manual Lane_move does not reset the lane extruder current back to print_current
### What happened?

I had a bowden tube pop out and used the lane_move command to get things going again and then got warnings about an over heating stepper driver. It seems that BT_Lane_move does not reset back to print_current. I think just adding an appropriate `[SET_TMC_CURRENT]` to that macro will fix it but I know that won't get updated for everyone

### Version

DEV

### Relevant log output

```shell

```

ejsears:
## Major Changes in this PR

This pull request addresses a bug related to the print current setting during the LANE_MOVE macro operation. The main change ensures that after moving a lane, the print current is properly restored, preventing issues with subsequent operations.

Bug fix for lane movement:

* Added a call to `cur_lane.set_print_current()` in the `cmd_LANE_MOVE` method in `extras/AFC.py` to correctly restore the print current after a lane move.
* Documented the bug fix in the `CHANGELOG.md` under the "Fixed" section for the 2025-12-03 release.

## Notes to Code Reviewers

Closes #578 

I can't test this right now, so not sure :)

### DIFF
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -704,7 +704,6 @@ class afc:
         cur_lane.unit_obj.return_to_home()
         # Put CAM back to lane if it is loaded to the toolhead
         self.function.select_loaded_lane()
-        cur_lane.set_print_current() # Set current back to print current after lane move
 
     def _get_resume_speed(self):
         """
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -702,8 +702,9 @@ class afc:
         cur_lane.do_enable(False)
         self.current_state = State.IDLE
         cur_lane.unit_obj.return_to_home()
-        # Put CAM back to lane if its loaded to toolhead
+        # Put CAM back to lane if it is loaded to the toolhead
         self.function.select_loaded_lane()
+        cur_lane.set_print_current() # Set current back to print current after lane move
 
     def _get_resume_speed(self):
         """
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [2025-12-03]
+### Fixed
+- Fixes bug where print current is not correctly set back to correct value after using LANE_MOVE macro.
+
 ## [2025-11-25]
 ### Added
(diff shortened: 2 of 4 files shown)


jimmyjon711:
Looking through the code a little more, I think we need to make this a little smarter and only set current back to print current if it was originally set before doing the lane move.

Or we could always just set load current when ejecting and when doing a lane load, as currently AFC assumes that current is set  to non print current before doing these operations.

extras/AFC.py
```diff
@@ -702,8 +702,9 @@ def cmd_LANE_MOVE(self, gcmd):
         cur_lane.do_enable(False)
         self.current_state = State.IDLE
         cur_lane.unit_obj.return_to_home()
-        # Put CAM back to lane if its loaded to toolhead
+        # Put CAM back to lane if it is loaded to the toolhead
         self.function.select_loaded_lane()
+        cur_lane.set_print_current() # Set current back to print current after lane move
```
jimmyjon711: After thinking about this for a bit, I think this would be better to move into `self.function.selecct_loaded_lane()` function and only set print current back to the lane that's loaded into the toolhead. That way current is not left lower as currently AFC assumes that current is set to load current before doing eject/load/etc.

jimmyjon711 (changes requested):

jimmyjon711:
Results from my changes @ejsears 
```
Doing lane move for lane1 while lane2 loaded in toolhead, lane2 current is set to print current since its loaded into toolhead
08:41:23   3780.506 [AFC_stepper.py:_set_current():187]          - DEBUG:Running macro: SET_TMC_CURRENT STEPPER='lane1' CURRENT=0.8
08:41:24   3781.330 [AFC_stepper.py:_set_current():187]          - DEBUG:Running macro: SET_TMC_CURRENT STEPPER='lane2' CURRENT=0.6

Unsetting lane2 from toolhead so current goes back to 0.8
08:41:51   3808.767 [AFC_stepper.py:_set_current():187]          - DEBUG:Running macro: SET_TMC_CURRENT STEPPER='lane2' CURRENT=0.8
08:41:52   3809.113 [AFC_functions.py:unset_lane_loaded():438]   - INFO: Manually removing lane2 loaded from toolhead

Doing lane move for lane1, current is not set back
08:41:54   3811.670 [AFC_stepper.py:_set_current():187]          - DEBUG:Running macro: SET_TMC_CURRENT STEPPER='lane1' CURRENT=0.8
```

jimmyjon711 (approved):
LGTM and tested

PR #618, opened 2026-01-23
macros: poop: fix querying fan speed for toolchangers

kekiefer:
The global fan object is no longer available if a fan_generic fan gets associated with a particular tool.

This change checks if a toolchanger setup is available, and if it is, will query the current fan speed from a loaded tool, if available. If not available, it will fall back on the global fan object, first checking if it exists, and then otherwise the fan is assumed to be off.


jimmyjon711 (approved):
LGTM, tested and worked as expected

PR #677, opened 2026-03-17
Toolchanger: fix error initializing after restart with tool on the shuttle

kekiefer:
## Major Changes in this PR

During PREP, force a call to note_detect_change() to re-query the detection pins and update detected_tool. This way later calls to INITIALIZE_TOOLCHANGER will set the correct tool as active.

Prior to this change, INITIALIZE_TOOLCHANGER would report 'active None' even if a tool was on the shuttle.

Manual verification and unit tests

The sequence to reproduce the problem solved here:

1. Run the printer as per normal, AFC_SELECT_TOOL leaves a tool on the shuttle
2. Restart klipper
3. Home and initialize KTC
4. Run AFC_UNSELECT_TOOL

Observe that the tool is not removed from the shuttle, unless this fix is applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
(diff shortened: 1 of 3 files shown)


kekiefer:
After some conference with @jimmyjon711, it seems like this is better addressed in KTC, since it's really a bug there.

PR #678, opened 2026-03-18
feat: add NEW_EXTRUDER_TEMP to Tn and automatic temp drop

kekiefer:
## Major Changes in this PR

Add NEW_EXTRUDER_TEMP to Tn and automatic temp drop

To better support toolchangers, we want to be able to pass the next extruder temperature to AFC so that it can set the temperature for the next tool to be loaded before the load happens. We also take the opportunity to adjust the temperature of the old extruder if it is not the same one being used.

This is particularly important when doing print-by-object, because Orca does not emit temperatures for other extruders not involved in printing the first object in a way they can be looked up in print start.

## Notes to Code Reviewers

In order to use this, change filament gcode should now the following. Note that if not changed, operations remain the same as they were:

```
T[next_extruder] PURGE_LENGTH={flush_length} NEW_EXTRUDER_TEMP={new_filament_temp} 
```

The start macro pattern to load the initial tool and ensure it is up to temperature is now:

```
T{INITIAL_TOOL} NEW_EXTRUDER_TEMP={EXTRUDER_TEMP} # Load Initial Tool
M109 S{EXTRUDER_TEMP} # wait for extruder temp, in case the tool was already loaded
```

Unit tests added for new functionality. This setup is meant to be used with a toolchanger and slicer configured for single-extruder multi-material.

### DIFF
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -10,7 +10,7 @@ import traceback
 import inspect
 from configfile import error
 
-from typing import Dict, TYPE_CHECKING, Union, Any
+from typing import Dict, TYPE_CHECKING, Union, Any, Optional, Tuple
 
 if TYPE_CHECKING:
     from configfile import ConfigWrapper
@@ -233,6 +233,9 @@ class afc:
             "restore_extruder_temp_on_load_or_unload", False
         )  # Restore extruder target temp after tool load/unload when not printing
         self.lower_extruder_temp_on_change = config.getboolean('lower_extruder_temp_on_change', True)  # When False, AFC will not lower extruder temp during filament change if already above target - 5
+        self.toolchange_temp_drop: float = config.getfloat(
+            "toolchange_temp_drop", 0
+        )  # Degrees to drop the old extruder's temperature (no wait) after a successful toolchange when the extruder changes.
         self.load_to_hub            = config.getboolean("load_to_hub", True)        # Fast loads filament to hub when inserted, set to False to disable. This is a global setting and can be overridden at AFC_stepper
         self.disable_homing_check   = config.getboolean("disable_homing_check", False)# Disables homing check when doing toolchanges. Only use this if you are using a toolchanger and don't need to home to unload toolheads
         self.assisted_unload        = config.getboolean("assisted_unload", True)    # If True, the unload retract is assisted to prevent loose windings, especially on full spools. This can prevent loops from slipping off the spool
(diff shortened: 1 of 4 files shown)


extras/AFC.py
```diff
@@ -2308,22 +2330,35 @@ def _heat_next_extruder(self, wait=True):
...
             deadband = next_extruder.deadband
             self._wait_for_temp_within_tolerance(next_heater, set_temp, deadband)
 
-        return next_heater, set_temp
+        return next_extruder, set_temp
+
+    def _cooldown_last_extruder(self, last_extruder: AFCExtruder, is_infinite_runout: bool) -> None:
+        """
+        Cools down the last extruder by setting its temperature to the specified value.
+        This function retrieves the heater of the last extruder and sets its temperature to the specified value.
```
jimmyjon711, line 2367: Can you please comment method's params

extras/AFC.py
```diff
@@ -2290,7 +2312,7 @@ def _cmd_AFC_M109(self, gcmd, wait=True):
         pheaters.set_temperature(heater, temp, should_wait)
         self.logger.debug("Done setting temp")
 
-    def _heat_next_extruder(self, wait=True):
+    def _heat_next_extruder(self, wait=True, next_temp: Optional[float]=None) -> Tuple[AFCExtruder, float]|bool:
```
jimmyjon711: Please add param comments and return comment to this method

jimmyjon711 (changes requested):
Some minor nit picks, but I have been trying to keep everything documented with method comments including passed in parameters and return comments

extras/AFC.py
kekiefer: Done

extras/AFC.py
```diff
@@ -2043,27 +2052,39 @@ def cmd_CHANGE_TOOL(self, gcmd):
...
+            if infinite_runout:
+                cur_lane.status = AFCLaneState.LOADED
+            next_extruder_obj = result[0]
             target_temp = result[1]
 
+            # Now cool down the old extruder (self.current changes during the toolchange,
+            # so capture the reference here after heating is already queued).
+            if self.current is not None:
+                _last_lane = self.lanes.get(self.current)
+                if _last_lane is not None and _last_lane.extruder_obj.name != next_extruder:
```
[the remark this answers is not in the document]
kekiefer: This is a poor suggestion. The temp drop needs to have the next temperature in hand on toolchange, or the subsequent change following a drop will occur at the too-low temperature. Nak.

kekiefer:
(Most) changes implemented, one omission with comment. One of those notes highlighted some test coverage I missed so I added that too.

jimmyjon711 (approved):
LGTM, simple swap and providing new temp works

kekiefer:
Yesterday I ran a single-toolhead multi-filament print with different temperatures. Having the new temperature in place for pre-heating up to the next temperature worked really well. We can probably recommend the macro update for all situations moving forward.

PR #682, opened 2026-03-22
Fix tool change not unloading stale lane on restart

kekiefer:
## Major Changes in this PR

At the start of a print, if the initial tool/extruder is not the
primary tool, and it already had filament loaded, self.current comes
back None. No unload would occur during CHANGE_TOOL, and old filament
would still be loaded to the toolhead. After this the initial
toolhead gets selected, and load fails if it's a different lane.
    
This change extends the load logic right to ensure that there is no
stale lane already loaded in the toolhead at the point before the
destination extruder is loaded, to protect againt this and other
occurrences in the future that might require an unload prior to load
before proceeding.

Load a lane in a secondary toolhead, leave it loaded and restart klipper for good measure. Start a print with a different lane on that same toolhead. This should not result in a failure to load.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2026-03-21]
+### Fixed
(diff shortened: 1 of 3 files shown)


kekiefer:
The new logic was moved out `CHANGE_TOOL` and into `TOOL_LOAD` to cover more cases than than just toolchanger, for example simply calling `TOOL_LOAD` at the start of a print when you've already got something in the tool.

jimmyjon711 (approved):
Tested on U1, works and LGTM

PR #683, opened 2026-03-23
Update Poop.cfg

dragoonmc:
## Major Changes in this PR
Added pulling in Purge modifier and length addition variables 
lined up the variables sections 
Applied variables to the provided length
 
### DIFF
--- a/config/macros/Poop.cfg
+++ b/config/macros/Poop.cfg
@@ -12,23 +12,25 @@ gcode:
   {% if extruder and printer[_th_key] is defined %}
       {% set _ = vars.update(printer[_th_key]) %}
   {% endif %}
-  {% set purge_x, purge_y       = vars.purge_loc_xy         | default([-1,-1])  | map('float') %}
-  {% set purge_spd              = vars['purge_spd']         | default(6.5)*60   | float %}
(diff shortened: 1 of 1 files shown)


ejsears:
I don't see any types of breaking changes here at all. I'm not a huge fan of using the same variable twice, but it isn't like this is super complicated. 

@jimmyjon711 thoughts?

dragoonmc:
thats fine, I wasn't sure of syntax for modifying it off the rip in line 66

config/macros/Poop.cfg
```diff
@@ -62,6 +64,7 @@ gcode:
     
   {% if params.PURGE_LENGTH %}
     {% set purge_len = params.PURGE_LENGTH|float %}
+    {% set purge_len = purge_len * purge_length_modifier + purge_length_addition %}
```
jimmyjon711, line 67: @dragoonmc you mean for this to he here and not after this if/else on line 70/71?

dragoonmc: I assumed that these modifiers were only to apply to slicer provided lengths based on the notes in the docs and cfgs.

PR #754, opened 2026-07-11
Support tollhead runout protections in bypass mode

cdheiser:
## Major Changes in this PR

This adds support to enable on filament runnout detection on toolhead sensors in bypass mode.

* **Added `enable_runout_in_bypass` Option**: Introduces a new configuration parameter `enable_runout_in_bypass` (defaulting to `False`). When set to `True`, it   
  allows the toolhead filament sensor(s) to trigger a print pause if a runout occurs while printing in bypass/manual mode.
* **Robust State Guards for Bypass Runout**: Added safety guards to the toolhead runout handler to ensure we only pause during active bypass prints:
    * **`not self.afc.in_toolchange`**: Suppresses runout triggers during toolchange unloads (preventing false pauses).
    * **`not self.afc.error_state`**: Prevents duplicate pause commands during the asynchronous transition when the print is already pausing.
    * **`self.on_shuttle()`**: Ensures parked/inactive toolheads in multi-tool configurations do not trigger runout pauses while the active toolhead is printing. 

- Unit tests updated and ran
- Tested on a Box Turtle

- [x] Documentation updated in AT-Documentation repository: https://github.com/AFCProject/Documentation/pull/18

## Summary by CodeRabbit

- **New Features**
  - Added `enable_runout_in_bypass` configuration option to control whether toolhead sensor runout triggers pausing in bypass/manual mode (default: disabled).

- **Bug Fixes**
  - Prevented unexpected runout pausing in bypass/manual mode unless `enable_runout_in_bypass` is explicitly enabled during active printing.
  
- **Documentation**
  - Updated `CHANGELOG.md` and AFC configuration to document the new option and behavior.

- **Tests**
  - Expanded unit tests to cover runout handling in bypass/manual printing vs. toolchange and error/idle states.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2026-07-11]
+### Added
+- Added `disable_tool_runout_in_bypass` configuration option (defaults to `False`) to allow users to toggle toolhead filament sensor runout pausing when printing in bypass/manual mode.
+
+### Fixed
+- Fixed toolhead filament sensor runout pausing print when printing in bypass/manual mode (where no AFC lane is registered as loaded).
+
 ## [2026-07-03]
 ### Added
 - Defaulting `tool_max_unload_attempts` to zero in Snapmaker U1 AFC config file.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [2026-07-11]
 ### Added
-- Added `disable_tool_runout_in_bypass` configuration option (defaults to `False`) to allow users to toggle toolhead filament sensor runout pausing when printing in bypass/manual mode.
+- Added `enable_runout_in_bypass` configuration option (defaults to `False`) to allow users to enable toolhead filament sensor runout pausing when printing in bypass/manual mode.
 
 ### Fixed
-- Fixed toolhead filament sensor runout pausing print when printing in bypass/manual mode (where no AFC lane is registered as loaded).
+- Fixed unexpected toolhead filament sensor runout pausing when printing in bypass/manual mode by defaulting the behavior to disabled.
 
 ## [2026-07-03]
 ### Added
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -262,7 +262,7 @@ class afc:
         self.enable_assist_weight   = config.getfloat("enable_assist_weight",   500.0)
         self.enable_hub_runout      = config.getboolean("enable_hub_runout",    True)
         self.enable_tool_runout     = config.getboolean("enable_tool_runout",   True)
-        self.disable_tool_runout_in_bypass = config.getboolean("disable_tool_runout_in_bypass", False)
+        self.enable_runout_in_bypass = config.getboolean("enable_runout_in_bypass", False)
         self.debounce_delay         = config.getfloat("debounce_delay",         0.)
 
         self.td1_when_loaded        = config.getboolean("capture_td1_when_loaded", False)
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -262,6 +262,7 @@ class afc:
         self.enable_assist_weight   = config.getfloat("enable_assist_weight",   500.0)
         self.enable_hub_runout      = config.getboolean("enable_hub_runout",    True)
         self.enable_tool_runout     = config.getboolean("enable_tool_runout",   True)
+        self.disable_tool_runout_in_bypass = config.getboolean("disable_tool_runout_in_bypass", False)
         self.debounce_delay         = config.getfloat("debounce_delay",         0.)
 
         self.td1_when_loaded        = config.getboolean("capture_td1_when_loaded", False)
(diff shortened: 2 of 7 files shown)


extras/AFC_extruder.py
```diff
@@ -495,10 +496,16 @@ def _handle_toolhead_sensor_runout(self, state, sensor_name):
         :param sensor_name: Name of the triggering sensor ("tool_start" or "tool_end")
         """
         # Notify the currently loaded lane if filament is missing at toolhead
-        if not state and self.lane_loaded and self.lane_loaded in self.lanes:
-            lane = self.lanes[self.lane_loaded]
-            if hasattr(lane, "handle_toolhead_runout"):
-                lane.handle_toolhead_runout(sensor=sensor_name)
+        if not state:
```
jimmyjon711, line 501: @cdheiser Please update the unit test for this changed code, all branches should be covered with your new updates for this method.

extras/AFC.py
```diff
@@ -262,6 +262,7 @@ def __init__(self, config: ConfigWrapper):
         self.enable_assist_weight   = config.getfloat("enable_assist_weight",   500.0)
         self.enable_hub_runout      = config.getboolean("enable_hub_runout",    True)
         self.enable_tool_runout     = config.getboolean("enable_tool_runout",   True)
+        self.disable_tool_runout_in_bypass = config.getboolean("disable_tool_runout_in_bypass", False)
```
jimmyjon711: Default to `True` so that everything works like before this change and then users can disable this(`False`) to have the runout trigger when using bypass. We can just add this in the docs on the feature page to let users know that this functionality exists.

extras/AFC_extruder.py
```diff
@@ -495,10 +496,16 @@ def _handle_toolhead_sensor_runout(self, state, sensor_name):
...
+            if self.lane_loaded and self.lane_loaded in self.lanes:
+                lane = self.lanes[self.lane_loaded]
+                if hasattr(lane, "handle_toolhead_runout"):
+                    lane.handle_toolhead_runout(sensor=sensor_name)
+            elif not self.disable_runout_in_bypass and self.afc.function.is_printing():
```
jimmyjon711: I have been trying to move all the code slowly to have the `if` statements like below:
```suggestion
            elif (not self.disable_runout_in_bypass
                  and self.afc.function.is_printing()):
```
Don't apply this change from github as I might have not gotten the spaces correct.

CHANGELOG.md
jimmyjon711, line 1: Please make sure your editor does not change the line endings to CRLF, line endings for this file should be LF

extras/AFC_extruder.py
jimmyjon711, line 494: Comment should be updated to inform about new updated.

jimmyjon711 (changes requested):
Thanks for submitting this PR, please see requested changes.

extras/AFC.py
cdheiser: Changed this to enable_tool_runout_in_bypass and default to false so the logic is clearer.

extras/AFC_extruder.py
cdheiser: Tweaked the formatting with the rename of the option.

CHANGELOG.md
cdheiser: It looks like this is CRLF in github, not LF.  If I convert to LF, the diff is massive.

extras/AFC_extruder.py
cdheiser: updated tests

extras/AFC_extruder.py
cdheiser: done

jimmyjon711: @coderabbitai review

extras/AFC_extruder.py
```diff
@@ -491,14 +492,23 @@ def _handle_toolhead_sensor_runout(self, state, sensor_name):
...
-        if not state and self.lane_loaded and self.lane_loaded in self.lanes:
-            lane = self.lanes[self.lane_loaded]
-            if hasattr(lane, "handle_toolhead_runout"):
-                lane.handle_toolhead_runout(sensor=sensor_name)
+        if not state:
+            if self.lane_loaded and self.lane_loaded in self.lanes:
+                lane = self.lanes[self.lane_loaded]
+                if hasattr(lane, "handle_toolhead_runout"):
+                    lane.handle_toolhead_runout(sensor=sensor_name)
+            elif (self.enable_runout_in_bypass
```
jimmyjon711: Please add checking that bypass is enabled with `self.afc.get_bypass_state()`, this is just more as a sanity check so it does not trigger other times if `self.enable_runout_in_bypass` is enabled and bypass is not enable.

jimmyjon711 (changes requested):
@cdheiser Additional change that I didn't catch last time. Also whats your discord name so I can credit you.

extras/AFC_extruder.py
cdheiser: I'm happy to add it (and that's what I originally coded up) but there's a scenario that I was concerned with:  physical bypass sensor.  If a user configures a physical bypass switch and disables pause on runout as the documentation suggests, then won't we fail to pause when the toolhead sensor trips because bypass mode won't be enabled anymore?  Maybe that means I should name this option "always_enable_tollhead_runout"?

jimmyjon711: I think the logic you are talking about should be added into the `lane.handle_toolhead_runout` method(I have TODOs there just haven't fully implemented it yet). But for this I think it makes sense to add the `self.afc.get_bypass_state()` guard since this needs to trigger only when runout is enabled since AFC does not look at the lane logic for this.

jimmyjon711: oh actually I just reread and I understand what your talking about now... hmmmmmm. Other problem is a user cant just always enable the enable_runout_in_bypass, cause this can also cause problems when not in bypass mode, hence why the `get_bypass_state()` check is needed.

cdheiser: Ahh I see your point now as I haven't run a print in non-bypass mode yet (as I'm printing a bunch of carbon fiber at the moment still).  Without the check, we'll pause the print on a lane change when we're not in bypass mode.

I updated the logic to check  not self.afc.in_toolchange . I also added guards for  not self.afc.error_state   
  │ (to prevent duplicate pause commands during async transitions) and  self.on_shuttle()  (to prevent inactive toolheads on multi-tool setups from triggering pauses).

I beleive this will now behave properly but will need to verify with a multi-color print on my BT which I'll do today.

cdheiser: I've confirmed that on my BT, this now proplery prevents freakout on a lane change.

jimmyjon711 (approved):
Looks good to me me, thanks again for your contribution.

PR #758, opened 2026-07-13
Add Duet 2 WiFi (SAM4E8E) pin mapping for BoxTurtle

rescosta:
## Summary

Adds a community MCU pin mapping (`community_mods/mcu/Duet2WiFi.cfg`) for running BoxTurtle on a **Duet 2 WiFi** (SAM4E8E, v1.00-1.04) instead of the official AFC-Lite board. Follows the AFC-X.cfg convention already used in this folder.

- Steppers use the on-board TMC2660 SPI drivers (E1/X/Y/Z connectors for lanes 1-4)
- N20 respoolers are driven through the 50-pin expansion connector into two TB6612FNG modules
- Hub, LOAD/PREP sensors, TurtleNeck buffer, RGB LEDs, and fan control all use free on-board/expansion pins
- Same alias interface as `AFC_Lite.cfg`, so the rest of a standard `AFC_Turtle_1.cfg` works with only the driver section changed to `[tmc2660]` (documented in the file header)

All pin assignments were cross-checked against the official Duet firmware source (`Pins_DuetNG.h`) and this mapping has been running on a physical 4-lane BoxTurtle build.

## Test plan

- [x] All 4 lanes homed/tested individually (`DUMP_TMC` clean on all drivers)
- [x] PREP/LOAD sensors verified per lane with real filament
- [x] TurtleNeck buffer sensors verified
- [x] Fan control (on-board temp-based on/off) verified in Mainsail
- [ ] Full multi-color print (pending FilamATrix toolhead cutter installation)


jimmyjon711 (approved):
LGTM, thanks for the contribution and all the comments in your mcu config file. I changed to merging into DEV branch instead of main.

PR #812, opened 2026-07-28
Save weight data to lane_data

### LINKED ISSUE #805: [Bug]: SET_WEIGHT never publishes weight to lane_data
### What happened?

`SET_WEIGHT` is the only one of the spool-metadata setters that never pushes to the Moonraker `lane_data` record.

`extras/AFC_spool.py` `cmd_SET_WEIGHT` (v1.2.0 `a06f14d`):

```python
        cur_lane = self.afc.lanes[lane]
        cur_lane.weight = weight
        self.afc.save_vars()
```

`SET_COLOR` and `SET_MATERIAL` both call `send_lane_data()`; this one only does `save_vars()`. This matters more since v1.2.0, because `send_lane_data()` now *includes* `weight` in the payload (`AFC_lane.py`, added in 74aa27b). So the field exists in the record but the command that sets it never publishes it — the value only reaches Moonraker if some unrelated command (`SET_COLOR`, `SET_MATERIAL`, `SET_MAP`, `AFC_SET_SPOOL_TEMP`, `set_spoolID`) happens to trigger a push afterwards.

Concretely: set a lane's weight via `SET_WEIGHT` and nothing else, and any consumer of `lane_data` keeps showing the old weight indefinitely.

Related, same area: `clear_lane_data()` has no `weight` key, so clearing a lane leaves a stale weight behind in the record.

I expected `SET_WEIGHT` to publish like its siblings.

### Version

### Relevant log output

```shell
n/a - behavioural, visible via GET /server/database/item?namespace=lane_data&key=lane1
```

ejsears:
## Major Changes in this PR

Use the same thing we do in other places to make sure we save the weight to `lane_data`.

Closes #805 

## Notes to Code Reviewers

Claude wrote the tests. I broke the tests to make sure they are actually doing what we expect.

New tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearing lane data now resets both the spool assignment and stored weight.
  * Setting a lane’s weight now immediately updates the lane data.

* **Tests**
  * Added coverage to verify weight updates are sent and fully cleared as expected.

### DIFF
--- a/extras/AFC_lane.py
+++ b/extras/AFC_lane.py
@@ -1773,7 +1773,8 @@ class AFCLane:
                     "td"             : "",
                     "lane"           : self.lane_index,
                     "extruder_index" : self.lane_extruder_index,
-                    "spool_id"       : None
+                    "spool_id"       : None,
+                    "weight"         : 0
                 }
             }
             self.afc.moonraker.send_lane_data(lane_data)
--- a/extras/AFC_spool.py
+++ b/extras/AFC_spool.py
@@ -255,6 +255,7 @@ class AFCSpool:
             return
         cur_lane = self.afc.lanes[lane]
         cur_lane.weight = weight
+        cur_lane.send_lane_data()
         self.afc.save_vars()
 
     cmd_SET_MATERIAL_help = "Sets filaments material for a lane"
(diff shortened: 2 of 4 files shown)


jimmyjon711 (approved):
LGTM

PR #813, opened 2026-07-29
fix(AFC_unit): use getboolean for remember_spool config

### LINKED ISSUE #806: [Bug]: unit-level remember_spool uses config.get, so an explicit False inherits as True
### What happened?

Unit-level `remember_spool` is read with `config.get` rather than `config.getboolean`, so an explicitly-configured `False` is inherited by lanes as `True`.

`extras/AFC_unit.py` (v1.2.0 `a06f14d`):

```python
self.remember_spool = config.get('remember_spool', False)   # returns a STRING when the option is present
```

Lane level is correct (`extras/AFC_lane.py`, `config.getboolean('remember_spool', None)`), and inheritance does:

```python
if self.remember_spool is None:
    self.remember_spool = bool(self.unit_obj.remember_spool)
```

`bool("False")` is `True` in Python, so a user who writes `remember_spool: False` in `[AFC_BoxTurtle ...]` gets lanes that *do* remember — the opposite of what they asked, with no warning.

It's latent rather than universal: when the option is absent, `config.get` returns the bool default `False` and everything behaves. It only bites on an explicit unit-level setting, and only the "off" spelling, which is also the spelling nobody tests because it looks like the default.

I confirmed this by inspection rather than observation — my own config omits the option, so my lanes correctly report `remember_spool: false`. Flagging it because the failure mode is silent and inverted.

### Version

### Relevant log output

```shell
python3 -c 'print(bool("False"))'  ->  True
```

abhi-0203:
## Summary

Closes #806 — Unit-level `remember_spool` is read with `config.get()` instead of `config.getboolean()`, so an explicitly-configured `False` is returned as the string `"False"` and inherited by lanes as `True` via `bool("False")`.

## Problem

In `AFC_unit.py`:
```python
self.remember_spool = config.get('remember_spool', False)  # returns string "False" when present
```

Lane-level correctly uses `config.getboolean('remember_spool', None)`, and inheritance does:
```python
if self.remember_spool is None:
    self.remember_spool = bool(self.unit_obj.remember_spool)  # bool("False") == True
```

So setting `remember_spool = False` at unit level silently enables it at lane level.

## Fix

Replace `config.get()` with `config.getboolean()` to properly parse the boolean value, matching the pattern already used in `AFC_lane.py`.

## Validation

- Verified `bool("False")` evaluates to `True` in Python (the bug)
- Verified `config.getboolean` returns actual `False` for `"False"` string values
- One-line fix, consistent with existing lane-level pattern

## Summary by CodeRabbit

* **New Features**
  * AFC status and AFC webhooks now expose the AFC version to help verify the running release.
* **Bug Fixes**
  * Lane clearing now resets lane weight (along with spool ID).
  * Setting lane weight now immediately propagates updated lane data before saving.
  * Buffer configuration replacement is more precise (only replaces on exact header matches).
  * Safer copy prompts in non-interactive runs now skip cleanly instead of misbehaving.
* **Tests**
  * Added unit tests covering AFC version reporting and webhook status payload shape, plus lane weight and SET_WEIGHT behavior.

### DIFF
--- a/extras/AFC_unit.py
+++ b/extras/AFC_unit.py
@@ -77,7 +77,7 @@ class afcUnit:
         self.extruder                    = config.get("extruder", None)                                      # Extruder name(AFC_extruder) that belongs to this unit, can be overridden in AFC_stepper section
         self.buffer_name                 = config.get('buffer', None)                                        # Buffer name(AFC_buffer) that belongs to this unit, can be overridden in AFC_stepper section
 
-        self.remember_spool              = config.get('remember_spool', False)                               # Turns on/off ability to remember last ejected spool values for all lanes in this unit, can be overridden in AFC_stepper section
+        self.remember_spool              = config.getboolean('remember_spool', False)                               # Turns on/off ability to remember last ejected spool values for all lanes in this unit, can be overridden in AFC_stepper section
         # LED SETTINGS
         # All variables use: (R,G,B,W) 0 = off, 1 = full brightness. Setting value here overrides values set in AFC.cfg file
         self.led_fault                   = config.get('led_fault', self.afc.led_fault)                       # LED color to set when faults occur in lane


jimmyjon711:
@abhi-0203 did you read my reply on your other PR https://github.com/AFCProject/AFC-Klipper-Add-On/pull/809#pullrequestreview-4799302283

Please use our PR template, stop having AI generated it.

PR #827, opened 2026-08-09
(fix): Guard handle_prep_runout against cross-lane PREP concurrency (Fixing Issue #826)

### LINKED ISSUE #826: [Bug]: MCU shutdown "Timer too close" when PREP sensor releases while another lane is still loading
### What happened?

I have a Box Turtle with 4 lanes hooked up to a Duet 2 SAM4E8E as the hub MCU. Under host I/O load (I forced this with a `dd` loop saturating the disk), if I release a lane's PREP sensor while that lane — or a different one — still has an active load cycle running, the hub MCU shuts down with "Timer too close". Sometimes it happens within a second of touching the sensor.

I dug into it and think I found the cause. `handle_prep_runout()` in `AFC_lane.py` reacts to the PREP switch going back to rest by calling `set_unloaded()`, which changes the lane's state and fires an LED update. There's no check anywhere for whether a PREP cycle (`prep_callback` → `prep_load` → `do_homing_move()`/`drip_move()`) is still actively running for that lane, or for a different one, at that exact moment. When it collides with the real-time step scheduling of an in-flight homing move, it looks like it trips Klipper's trsync watchdog and the MCU goes down.

Early on I was actually looking at the press itself — tried different speeds, disabled synchronous `save_vars()`, that kind of thing — since that's the moment the motor visibly starts moving and the crash follows shortly after. It was only after ruling all of that out that I started considering the release as the actual trigger, and once I tested specifically for that, the pattern became clear.

I've seen a few similar reports from other users on Discord (different hardware than mine, same exact "Timer too close" symptom), and none of them draw a line between press and release either — they all describe it generically as "crashes while loading a lane". I think that's the same thing that slowed me down: the press is the obvious thing to suspect, since that's when the motor visibly starts turning, so nobody stopped to isolate the release as the real trigger.

I confirmed the pattern by hand: holding PREP and releasing it while the cycle is still running reliably crashes it; releasing after the cycle already finished on its own never does. It also happens across lanes — releasing lane A while lane B is still mid-cycle triggers it too.

I put together a fix that adds two cross-lane checks before `handle_prep_runout` acts on that edge (how many lanes currently have a PREP cycle in flight, and the timestamp of the last PREP edge seen on any lane), so it holds off the state change/LED update while something else is still going, without touching the mid-print runout/pause path at all — that one needs to react immediately, so it's never gated. Tested repeatedly on real hardware and the fast, reliable crash is gone. It's been running for a while now in normal use (no forced `dd`) and hasn't thrown a single new "Timer too close" so far. Also added unit test coverage for it and ran the full suite plus the klippy integration tests — everything green.

I have a PR ready with this fix if it's useful.

Investigation and fix drafted with the help of Claude, but the actual breakthrough — realizing the trigger was the sensor releasing rather than the press — came from my own testing on the printer, not something Claude suggested. Tested on real hardware and edited by me throughout.

### Version

v1.2.0-1845-a06f14d

### Relevant log output

```shell
MCU 'Turtle_1' shutdown: Timer too close
This often indicates the host computer is overloaded. Check
for other processes consuming excessive CPU time, high swap
usage, disk errors, overheating, unstable voltage, or
similar system problems on the host computer.
Printer is shutdown
```

rescosta:
## Major Changes in this PR

Releasing the PREP sensor mutates lane state and also updates the status LED on the hub MCU. When that happens while the same (or another) lane still has an active homing move in flight, the LED update competes with real-time step scheduling on that MCU, delaying the sync confirmation the Klipper `trsync` watchdog expects, which trips a "Timer too close" shutdown.

- `handle_prep_runout()` in `extras/AFC_lane.py` now checks whether any lane has an active PREP cycle in flight before acting on a runout edge (releasing the sensor), instead of acting unconditionally. Each lane already tracked its own active/inactive state; a new method,`_any_lane_prep_active()`, checks that across every lane.
- If PREP activity is in or near flight, `handle_prep_runout()` doesn't drop the release, it defers: schedules a recheck instead of acting right away. Each lane only keeps one recheck pending at a time.
- The mid-print runout/pause behavior isn't touched by this guard, it always reacts immediately, delaying it would be a real safety regression.
- No behavior change for anyone who never hits this race, same defaults, same result.

## Notes to Code Reviewers

- `PREP_GUARD_WINDOW = 2.0` (seconds) was chosen empirically from live reproduction testing, not derived analytically, open to discussion if a different value makes more sense.
- Started with a shared counter (`prep_active_count`) across all lanes. Switched to the boolean approach above per review feedback, since each lane's own active/inactive flag was already the real source of truth and a separate mirrored counter wasn't needed.
- Under conditions well beyond disk I/O saturation alone — a deliberately forced sequence of pressing/releasing PREP across all 4 lanes, repeated over several attempts, at a pace far above any normal use of the system — the crash was still reproduced once in testing. Not observed under any realistic usage pattern.

- Live reproduction on real hardware (Voron 2.4, Box Turtle, 4 lanes, Duet 2 SAM4E8E hub MCU), with a disk I/O stress loop running in parallel to simulate host overload.
- Before the fix: under that same I/O stress, reliable crash releasing PREP mid-cycle, same lane or cross-lane.
- After the fix: repeated the same test under stress, no crash matching real usage patterns. Running in normal use since, no new "Timer too close".
- `ruff check .`: clean. `pytest tests/`: full suite passing (3268 tests). `pytest tests/klippy/`: integration suite passing (7 passed, 2 xfailed as expected).

## PR Checklist

---

The investigation and this fix were done with the help of Claude, but the actual breakthrough, realizing the trigger was the sensor releasing rather than the press, came from my own testing on the printer, not something Claude suggested. Tested on real hardware and reviewed by me throughout.

## Bug Fixes

* Reduced “Timer too close” MCU shutdowns when a PREP sensor releases during an active load cycle.
* Added a two-second safety window around recent PREP activity.
* Coordinated PREP activity across lanes to prevent overlapping operations.
* Deferred non-printing unload actions until PREP activity and recent sensor activity have cleared.
* Preserved immediate runout handling during printing and improved recovery when preparation is interrupted or debounce controls are unavailable.

### DIFF
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -108,7 +108,6 @@ class afc:
         self._last_td1_query:float    = 0
         self.lane_data_enabled  = False
         self.prep_done          = False         # Variable used to hold of save_vars function from saving too early and overriding save before prep can be ran
-        self.prep_active_count       = 0    # Lanes with an active PREP cycle right now
         self.last_prep_activity_time = 0.0  # eventtime of the last PREP edge on any lane
         self.in_print_timer     = None
         self.activate_cb_done = True
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -108,6 +108,13 @@ class afc:
         self._last_td1_query:float    = 0
         self.lane_data_enabled  = False
         self.prep_done          = False         # Variable used to hold of save_vars function from saving too early and overriding save before prep can be ran
+        # --- TTC (Timer Too Close) fix: cross-lane PREP runout guard ------------------------
+        # See handle_prep_runout() in AFC_lane.py for the full explanation. These two fields
+        # let that function detect PREP activity on ANY lane (not just its own) before acting
+        # on a runout edge, closing a race that could shut down the hub MCU with "Timer too
+        # close" (fixed 2026-08-08).
+        self.prep_active_count       = 0    # How many lanes currently have an active PREP cycle
+        self.last_prep_activity_time = 0.0  # eventtime of the last PREP edge seen on any lane
         self.in_print_timer     = None
         self.activate_cb_done = True
         self.db_backup          = False
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -108,13 +108,8 @@ class afc:
         self._last_td1_query:float    = 0
         self.lane_data_enabled  = False
         self.prep_done          = False         # Variable used to hold of save_vars function from saving too early and overriding save before prep can be ran
-        # --- TTC (Timer Too Close) fix: cross-lane PREP runout guard ------------------------
-        # See handle_prep_runout() in AFC_lane.py for the full explanation. These two fields
-        # let that function detect PREP activity on ANY lane (not just its own) before acting
-        # on a runout edge, closing a race that could shut down the hub MCU with "Timer too
-        # close" (fixed 2026-08-08).
-        self.prep_active_count       = 0    # How many lanes currently have an active PREP cycle
-        self.last_prep_activity_time = 0.0  # eventtime of the last PREP edge seen on any lane
+        self.prep_active_count       = 0    # Lanes with an active PREP cycle right now
+        self.last_prep_activity_time = 0.0  # eventtime of the last PREP edge on any lane
         self.in_print_timer     = None
         self.activate_cb_done = True
         self.db_backup          = False
--- a/extras/AFC_lane.py
+++ b/extras/AFC_lane.py
@@ -1262,6 +1262,10 @@ class AFCLane:
                                 self.afc.error.AFC_error(f"Cannot load {self.name} spool while printer is actively moving or homing", False)
                                 return
 
+                            # Check to see if another lane already has a PREP cycle in flight
+                            if any(lane.prep_active for lane in self.afc.lanes.values() if lane is not self):
+                                return
+
                             # Calling common load function
                             self.unit_obj.prep_load(self)
 
@@ -1361,13 +1365,19 @@ class AFCLane:
         self.afc.save_vars()
 
     def _any_lane_prep_active(self) -> bool:
-        """True if any lane (this one or another) has a PREP cycle in flight right now."""
+        """
+        Check whether any lane currently has a PREP cycle in flight.
+
+        :return: True if any lane (this one or another) is mid-cycle right now
+        """
         return any(lane.prep_active for lane in self.afc.lanes.values())
 
-    def _recheck_prep_runout(self, eventtime):
+    def _recheck_prep_runout(self, eventtime: float) -> None:
         """
         Re-evaluate a PREP release that handle_prep_runout() deferred. Defers
         again if still blocked or the lane got filament back in the meantime.
+
+        :param eventtime: reactor time this recheck fired
         """
         if self.prep_state:
             self.prep_recheck_pending = False
--- a/extras/AFC_lane.py
+++ b/extras/AFC_lane.py
@@ -1237,72 +1237,79 @@ class AFCLane:
         self.prep_active = True
         self.afc.prep_active_count += 1   # TTC fix: mirror into the shared cross-lane counter
 
-        # Checking to make sure printer is ready and making sure PREP has been called before trying to load anything
-        for i in range(1):
-            # Hacky way for do{}while(0) loop, DO NOT return from this for loop, use break instead so that self.prep_state variable gets sets correctly
-            #  before exiting function
-            with self.mutex:
-                if (self.printer.state_message == 'Printer is ready'
-                    and self._afc_prep_done
-                    and self.status != AFCLaneState.TOOL_UNLOADING):
-                    # Only try to load when load state trigger is false
-                    if (self.prep_state
-                        and not self.raw_load_state):
-                        self.logger.debug(f"Prep: callback triggered {self.name}")
-                        # Checking to make sure last time prep switch was activated was less than 1 second, returning to keep is printing message from spamming
-                        # the console since it takes klipper some time to transition to idle when idle_resume=printing
-                        if delta_time < 1.0:
-                            break
+        # TTC fix: the increment above must always be undone, even if something in this block
+        # raises (prep_load()/prep_post_load()/TOOL_LOAD()/etc. are all reachable here and none
+        # of them are guaranteed not to throw). Without this, an unhandled exception would leave
+        # prep_active_count stuck positive forever, permanently blocking handle_prep_runout's
+        # non-printing set_unloaded() path for every lane, not just this one. save_vars() stays
+        # outside the try/finally on purpose: it must still run only on normal completion
+        # (fall-through/break), not on the early-return abort below or on a raised exception.
+        try:
+            # Checking to make sure printer is ready and making sure PREP has been called before trying to load anything
+            for i in range(1):
+                # Hacky way for do{}while(0) loop, DO NOT return from this for loop, use break instead so that self.prep_state variable gets sets correctly
+                #  before exiting function
+                with self.mutex:
+                    if (self.printer.state_message == 'Printer is ready'
+                        and self._afc_prep_done
+                        and self.status != AFCLaneState.TOOL_UNLOADING):
+                        # Only try to load when load state trigger is false
+                        if (self.prep_state
+                            and not self.raw_load_state):
+                            self.logger.debug(f"Prep: callback triggered {self.name}")
+                            # Checking to make sure last time prep switch was activated was less than 1 second, returning to keep is printing message from spamming
+                            # the console since it takes klipper some time to transition to idle when idle_resume=printing
+                            if delta_time < 1.0:
+                                break
 
-                        # Check to see if the printer is printing or moving, as trying to load while printer is doing something will crash klipper
-                        if self.afc.function.is_printing(check_movement=True):
-                            self.afc.error.AFC_error(f"Cannot load {self.name} spool while printer is actively moving or homing", False)
-                            self.prep_active = False
-                            self.afc.prep_active_count -= 1   # TTC fix: mirror exit here too
-                            return
+                            # Check to see if the printer is printing or moving, as trying to load while printer is doing something will crash klipper
+                            if self.afc.function.is_printing(check_movement=True):
+                                self.afc.error.AFC_error(f"Cannot load {self.name} spool while printer is actively moving or homing", False)
+                                return
 
-                        # Calling common load function
-                        self.unit_obj.prep_load(self)
+                            # Calling common load function
+                            self.unit_obj.prep_load(self)
 
-                        self.status = AFCLaneState.NONE
-                        self.logger.debug(f"Prep: Load Done-{self.name}")
+                            self.status = AFCLaneState.NONE
+                            self.logger.debug(f"Prep: Load Done-{self.name}")
 
-                        # Verify that load state is still true as this would still trigger if prep sensor was triggered and then filament was removed
-                        #   This is only really a issue when using direct_load and still using load sensor
-                        if (self.hub == 'direct_load'
-                            and self.prep_state):
-                            self.logger.debug(f"Prep: direct load logic-{self.name}-{self.hub}")
-                            if not self.afc.TOOL_LOAD(self):
-                                self.afc.afc_stats.increase_load_error_count(self.afc)
-                            self.afc.spool._set_values(self)
-                            self.logger.debug(f"Prep: direct load logic done-{self.name}-{self.hub}")
-                            break
+                            # Verify that load state is still true as this would still trigger if prep sensor was triggered and then filament was removed
+                            #   This is only really a issue when using direct_load and still using load sensor
+                            if (self.hub == 'direct_load'
+                                and self.prep_state):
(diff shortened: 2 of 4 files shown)


extras/AFC.py
```diff
@@ -108,6 +108,13 @@ def __init__(self, config: ConfigWrapper):
         self._last_td1_query:float    = 0
         self.lane_data_enabled  = False
         self.prep_done          = False         # Variable used to hold of save_vars function from saving too early and overriding save before prep can be ran
+        # --- TTC (Timer Too Close) fix: cross-lane PREP runout guard ------------------------
```
ejsears: Can probably remove this multi-line comment.

extras/AFC_lane.py
```diff
@@ -1332,6 +1346,34 @@ def handle_prep_runout(self, eventtime, prep_state):
                 else:
                     self._perform_pause_runout()
             elif not prep_state:
+                # --- TTC (Timer Too Close) fix ----------------------------------------------
```
ejsears: Same here with the comment, or at least slim it down quite a bit. @jimmyjon711 thoughts?

jimmyjon711: @ejsears @rescosta The one problem that I see here is that if someone ejected, didnt fully remove the filament and then inserted filament into another lane and then removed the ejected lane, the lane that was just ejected will not update its internal status and wont show unloaded properly.

extras/AFC.py
```diff
@@ -108,6 +108,13 @@ def __init__(self, config: ConfigWrapper):
...
+        # See handle_prep_runout() in AFC_lane.py for the full explanation. These two fields
+        # let that function detect PREP activity on ANY lane (not just its own) before acting
+        # on a runout edge, closing a race that could shut down the hub MCU with "Timer too
+        # close" (fixed 2026-08-08).
+        self.prep_active_count       = 0    # How many lanes currently have an active PREP cycle
```
jimmyjon711: So only one lane can run prep at a time, this would make more sense to make this into a boolean instead instead of a counter.

extras/AFC.py
```diff
@@ -108,6 +108,13 @@ def __init__(self, config: ConfigWrapper):
         self._last_td1_query:float    = 0
         self.lane_data_enabled  = False
         self.prep_done          = False         # Variable used to hold of save_vars function from saving too early and overriding save before prep can be ran
+        # --- TTC (Timer Too Close) fix: cross-lane PREP runout guard ------------------------
+        # See handle_prep_runout() in AFC_lane.py for the full explanation. These two fields
```
jimmyjon711: Yes I also agree with Eric, please limit the amount of AI comment spam. The agents file specifically say to keep comments short and succulent, no need to long comment blocks.

jimmyjon711 (changes requested):

extras/AFC.py
rescosta: Good question, and fair to push on it. The reason I went with a counter: PREP isn't instant, the press itself just kicks off a homing move (drip_move) that can take a few seconds to reach the LOAD sensor. That call blocks inside a loop that calls reactor.pause() repeatedly, and that's what lets other lanes run in between, it's not one Python callback blocking another, it's two lanes with a cycle physically in flight at the same time. Confirmed this two ways: holding all 4 lanes down at once, and also pressing them one after another without releasing the previous one, both times the motors kept spinning independently, none waited on the others.

A boolean breaks in exactly that overlap: lane A starts (True), lane B starts while A is still going (still True, no-op), B finishes first and sets it back to False, but A is still mid-move. If a runout lands right then, the guard reads False and lets it through, which is the same race we're trying to close. The counter avoids that because it tracks how many are active, not just whether any are.

jimmyjon711: Not sure how you are getting all lanes to do a homing move at once as I have not been able to replicate what you are saying, this can also cause TTCs. A user should only be able to load a single lane at a time, not multiple. This is why I suggested going to a boolean, but this global flag also needs to be added to line 1258 of AFC_lane.py

May I also kindly remind you about [this](<https://github.com/AFCProject/AFC-Klipper-Add-On/blob/main/AI_USAGE_POLICY.md#interacting-with-maintainers-should-be-human>) in our AI use policy.

rescosta: Fair to call that out, and I think I know exactly why you never reproduced it and I did.

The guard that's supposed to prevent this runs inside prep_callback itself, during loading, not printing, it's meant to stop lane B from starting while lane A is still moving. But it only tracks the main printer's motion state (is_printing(check_movement=True), which checks idle_timeout), it never sees PREP's own internal homing move, so it doesn't catch lane-vs-lane overlap during loading at all.

This isn't a forced edge case either, I can trigger this by pressing the PREP switches directly, no filament needed, and pressing all 4 lanes makes all 4 load motors spin. That's just normal usage on my printer, not something I had to fight to trigger. If you test the same way (press multiple PREP switches close together instead of one at a time with a wait), I think you'll be able to reproduce it too.

Agree on the line 1258 comment too, already fixed in a commit I haven't pushed yet (same finding CodeRabbit made, actually). Trimmed the comments as well, same fix. Pushing both shortly.

jimmyjon711: @rescosta What version of klipper are you currently using?

rescosta: Version: v0.13.0-730-g84104bbb

jimmyjon711: @rescosta Updated to the newest klipper and I am now able to trigger different lanes at the same time. But for now lets please move to using a boolean instead of a counter. I think being able to load multiple lanes at a time could be useful, but this is something that I think is out of the scope of this PR.

jimmyjon711:
@rescosta Please implement this nitpick comment from coderabbit for AFC_lane.py https://github.com/AFCProject/AFC-Klipper-Add-On/pull/827#pullrequestreview-4893337990

extras/AFC.py
rescosta: Will make the adjustments, you're right, thought it over and see what you meant about the boolean now. It's early morning here, working through it this morning and I'll push shortly after.

rescosta: I just pushed the changes, and before that I tested them physically on the BT, working as expected.

extras/AFC_lane.py
```diff
@@ -1232,70 +1236,74 @@ def prep_callback(self, eventtime: float, state: int) -> None:
...
+                        if (self.prep_state
+                            and not self.raw_load_state):
+                            self.logger.debug(f"Prep: callback triggered {self.name}")
+                            # Checking to make sure last time prep switch was activated was less than 1 second, returning to keep is printing message from spamming
+                            # the console since it takes klipper some time to transition to idle when idle_resume=printing
+                            if delta_time < 1.0:
+                                break
+
+                            # Check to see if the printer is printing or moving, as trying to load while printer is doing something will crash klipper
+                            if self.afc.function.is_printing(check_movement=True):
```
jimmyjon711: Please add the `_any_lane_prep_active` method check here as well to limit only one lane being active at a time.

extras/AFC_lane.py
```diff
@@ -1332,11 +1344,45 @@ def handle_prep_runout(self, eventtime, prep_state):
...
+                            self._recheck_prep_runout,
+                            self.reactor.monotonic() + self.PREP_GUARD_WINDOW,
+                        )
+                    return
                 # Filament is unloaded
                 self.set_unloaded()
 
         self.afc.save_vars()
 
+    def _any_lane_prep_active(self) -> bool:
```
jimmyjon711, line 1365: Please update comment to match current style and this also needs the return documented

extras/AFC_lane.py
```diff
@@ -1332,11 +1344,45 @@ def handle_prep_runout(self, eventtime, prep_state):
...
+        """True if any lane (this one or another) has a PREP cycle in flight right now."""
+        return any(lane.prep_active for lane in self.afc.lanes.values())
+
+    def _recheck_prep_runout(self, eventtime):
```
jimmyjon711: Please add typing and parameter needs to be added in method comment for `eventtime`

jimmyjon711 (changes requested):

extras/AFC_lane.py
rescosta: I tried a simpler approach for that point in the code: check that it's the same lane and let it through, because at that point it was blocking every lane that came through. Is this something like what you had in mind? Before you asked for this, I had already been thinking about suggesting it myself. I ran some tests on the BT forcing a rapid sequence across lanes, and it triggered a Timer Too Close from the homing load.

jimmyjon711: Goal is to only allow on lane to be loaded at a time, if a user tries to load in another spool while the previous lane is still loading it should not activate another lanes motor.

rescosta: Either way, even with the code validated, I still want to stress-test it more here on my BT.

extras/AFC_lane.py
```diff
@@ -1232,70 +1236,78 @@ def prep_callback(self, eventtime: float, state: int) -> None:
...
+                            if delta_time < 1.0:
+                                break
+
+                            # Check to see if the printer is printing or moving, as trying to load while printer is doing something will crash klipper
+                            if self.afc.function.is_printing(check_movement=True):
+                                self.afc.error.AFC_error(f"Cannot load {self.name} spool while printer is actively moving or homing", False)
+                                return
+
+                            # Check to see if another lane already has a PREP cycle in flight
+                            if any(lane.prep_active for lane in self.afc.lanes.values() if lane is not self):
```
jimmyjon711: @rescosta This check should be added to the if statement on line 1261, that way a message is displayed to the user before returning.

jimmyjon711 (changes requested):

jimmyjon711:
@coderabbitai full review

extras/AFC_lane.py
```diff
@@ -1214,6 +1216,8 @@ def prep_callback(self, eventtime: float, state: int) -> None:
         :param state: 1 if the prep sensor is now triggered (filament present), 0 otherwise
         """
         self.prep_state = bool(state)
+        # Record every PREP edge (any lane) for handle_prep_runout's guard
+        self.afc.last_prep_activity_time = eventtime
```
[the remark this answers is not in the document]
jimmyjon711: @rescosta I agree with this comment, probably should update this time when a prep actually starts. So moving to after 1264 I think would be the right thing to do.

rescosta:
After these tests where homing became protected against TTC and stopped showing that problem, another issue came up during stress testing here, a completely separate mechanism from this PR's PREP guard. It probably never showed up before because the homing crash was happening first. It's in the espooler's stats callback, which runs every 30 seconds with all lanes firing together. I saw it in the logs when the new TTC happened, but I'm not sure if this would be something for this same PR. I was thinking of just adding a delay between each lane's calls so they don't coincide, and testing that before doing anything else. 

[afc_espooler_ttc_raw_1.log](https://github.com/user-attachments/files/31066453/afc_espooler_ttc_raw_1.log)
[afc_espooler_ttc_raw_2.log](https://github.com/user-attachments/files/31066455/afc_espooler_ttc_raw_2.log)

jimmyjon711:
> After these tests where homing became protected against TTC and stopped showing that problem, another issue came up during stress testing here, a completely separate mechanism from this PR's PREP guard. It probably never showed up before because the homing crash was happening first. It's in the espooler's stats callback, which runs every 30 seconds with all lanes firing together. I saw it in the logs when the new TTC happened, but I'm not sure if this would be something for this same PR. I was thinking of just adding a delay between each lane's calls so they don't coincide, and testing that before doing anything else.
> 
> [afc_espooler_ttc_raw_1.log](https://github.com/user-attachments/files/31066453/afc_espooler_ttc_raw_1.log) [afc_espooler_ttc_raw_2.log](https://github.com/user-attachments/files/31066455/afc_espooler_ttc_raw_2.log)

@rescosta This is out of the scope from this PR in my opinion, also I think I need to understand the new klipper changes and how it affects our `is_printing` check since this should have not been sent up to moonraker if a lanes motor is actively moving. This is better off creating a new issue and we can fix it in a different PR.

extras/AFC_lane.py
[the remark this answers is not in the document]
jimmyjon711: @coderabbitai review

rescosta:
@jimmyjon711  During the tests I ran today, the callback fires right after the initial test on the 4 lanes, 30s later. Since there's no more movement after that, it doesn't happen again. But nothing prevents the callback from firing during homing. So I saw the TTC wasn't the spool callback itself, but it hitting during homing. I fixed that by blocking the callback during homing, but I still need to test more here. So I agree with you about not including this in this PR. Let's take it one thing at a time.

extras/AFC_lane.py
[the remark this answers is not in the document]
jimmyjon711: @coderabbitai this should have been fixed in https://github.com/AFCProject/AFC-Klipper-Add-On/pull/827/commits/9245177f8e09d3119048c5aad00e7884bbe5d0a6

jimmyjon711 (approved):
Changes LGTM, tested with triggering a prep load and triggered other lanes. Kept getting cannot prep lane message as it should be.

@rescosta I am going to merge this PR, could you please create a issue for the other thing that you found for the espooler and how to replicate.

rescosta:
@jimmyjon711 Created issue #831 with the details.

PR #830, opened 2026-08-13
(fix): Report LANE_UNLOAD refusals as warnings so clients can see them

prestonbrown:
## Minor Change in this PR

LANE_UNLOAD refuses in two cases, but feedback isn't provided to consumers except the console:

- if lane is the one loaded in the toolhead: logger.info only, so it never reaches message_queue
- if a standalone extruder with no lane loaded: logs nothing, if/elif has no handler for this

Both situations now emit at logger.warning.  Therefore they end up in the message queue and are
retrievable from AFC.message via get_status.  That's it.

This matters because LANE_UNLOAD has no success/error feedback. Moonraker returns ok whether the
eject happened or not, so from a client's side, you can't tell if it worked or failed. The only
refusal a UI can properly observe is the is_printing one, which goes through AFC_error. This makes
(for example) a UI "Eject" button appear to do nothing at all (or at least, the client has to assume
success), which is a sub-optimal UX.

The toolhead message got modified to hint the user to "Run TOOL_UNLOAD first." If you don't like that
we can drop it but it seems it would help nudge users in the right direction.

## Notes to Code Reviewers

LANE_UNLOAD had no test class, so I added one covering all four cases rather than only the two lines
I touched. That's most of the diff.

The else case is reachable when it's a standalone extruder, the lane being ejected isn't the currently
loaded one, and lane_loaded is false. I'll have to defer to you about whether this should actually
eject the lane rather than gently fail, since nothing is loaded anywhere. Behavior is unchanged, only
the warning was added.

Obviously I did all of this to support HelixScreen UI/UX.

18 new tests in `tests/test_AFC.py::TestLaneUnload`, covering all four arms of the if/elif chain.

- `ruff check .` clean
- full unit suite green, 3259 tests
- `--cov-branch` shows no missing or partial branches in LANE_UNLOAD
- confirmed the tests fail against the old code, not just pass against the new: reverting `warning`
  back to `info` fails `test_lane_in_toolhead_refuses_and_warns`, and deleting the `else` fails
  `test_standalone_without_lane_loaded_warns`

Klippy integration tests not run. Nothing here touches the gcode surface, and running them needs
klipper and kalico checkouts plus a firmware build on my end. Happy to if you want them.

---

Per AI_USAGE_POLICY.md: code and tests were drafted with Claude, then reviewed, run, and edited by
me. This description is mine.

## Summary by CodeRabbit

* **Bug Fixes**
  * `LANE_UNLOAD` now warns when unloading is refused because the lane is loaded in the toolhead.
  * Attempts to unload when no lane is loaded on a standalone extruder now produce a clear warning.
  * Lane state and temporary ejecting conditions are handled correctly during unloading.

* **Documentation**
  * Added changelog details describing the updated warning behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2026-08-13]
+### Fixed
+- `LANE_UNLOAD` now reports its refusals as warnings instead of console-only messages, so they reach
+  the message queue and show up in Mainsail/Fluidd and other clients. Previously, ejecting a lane that
+  was loaded in the toolhead logged to the console only, and a standalone extruder with no lane loaded
+  logged nothing at all - in both cases the command returned normally, so a UI could not tell a refused
+  eject from a completed one.
+
 ## [2026-08-09]
 ### Fixed
 - Fixed a "Timer too close" MCU shutdown that could occur when the PREP sensor releases while a
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -1365,6 +1365,15 @@ class afc:
         self.LANE_UNLOAD( cur_lane )
 
     def LANE_UNLOAD(self, cur_lane: AFCLane):
+        """
+        Ejects a lane's filament back out of the unit, when the lane is not in use.
+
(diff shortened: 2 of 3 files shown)


jimmyjon711:
@prestonbrown Can you please rebase DEV and fix the merge conflict in CHANGELOG.md, thanks

prestonbrown:
Rebased on DEV and resolved the CHANGELOG.md conflict - should be clean now.

jimmyjon711 (approved):
LGTM

PR #833, opened 2026-08-16
Publish vendor, filament name and initial weight in lane_data (#808)

### LINKED ISSUE #808: [Enhancement]: publish vendor / filament_name / initial_weight in lane_data and status
### Is your feature request related to a problem?

`lane_data` is the interop surface slicers and UIs read, but it can't currently describe a spool's *brand*, so anything consuming it has to fall back to a second source or just not show vendor at all.

The interesting part is that v1.2.0 already fetches the data — it just doesn't publish it. In `AFC_spool.set_spoolID()`:

```python
cur_lane.filament_name = self._get_filament_values(result['filament'], 'name')
cur_lane.spool_vendor  = self._get_filament_values(result['filament']['vendor'], 'name', default="")
...
cur_lane.espooler.espooler_values.full_weight = self._get_filament_values(result, 'initial_weight', default=1000)
```

So vendor, filament name and full weight are all already on the lane object after a Spoolman lookup.

### Describe the solution you'd like

Two small additions:

1. **Include them in `send_lane_data()`** (`AFC_lane.py`), alongside the `weight` and `extruder_index` that were added in v1.2.0:
   - `vendor` (from `spool_vendor`)
   - `filament_name`
   - `initial_weight` (from `espooler_values.full_weight`)

2. **Expose `spool_vendor` in `get_status()`.** `filament_name` and `initial_weight` made it into the live status dict in v1.2.0; `spool_vendor` is currently a lane attribute only, so it's unreachable for anything reading status.

Nothing new to fetch or store — it's publishing values the lane already holds.

### Describe alternatives you've considered

A client can query Spoolman itself for the vendor given `spool_id`, which is what we do today, but that's a second round trip for data AFC already has, and it doesn't help a lane with manually-entered metadata and no Spoolman link.

We could also keep this in our own private Moonraker namespace, which is what we're about to do as a stopgap — but that fragments lane metadata across two places and doesn't help OrcaSlicer or anyone else reading `lane_data`.

### Additional context

Happy to send a PR if you'd like it — it looks like a handful of lines in `send_lane_data()` plus one entry in the status dict.

Context: HelixScreen shows per-lane brand/vendor in its AMS UI. `lane_data` gives us colour, material, temps and `spool_id`, but no vendor, so we currently reconcile against Spoolman separately. Filed alongside #803-#806.

prestonbrown:
## Major Changes in this PR

Closes #808.

`lane_data` is what slicers and UIs read, but it could not describe a spool's brand, so a
consumer could see that a lane held PLA and nothing about whose PLA it was. AFC already
fetches vendor, filament name and initial weight during a Spoolman lookup and attaches them
to the lane; they were just never published.

- `send_lane_data` now emits `vendor_name`, `name` and `initial_weight`
- `clear_lane_data` clears them alongside the fields it already reset
- `get_status` gains `spool_vendor`, the only one of the three that was missing from the
  live status dict

The `lane_data` keys are the ones already established for that shared namespace,
`vendor_name` and `name`, per your note on #808. `get_status` reports the vendor under
this class's own attribute name, `spool_vendor`, alongside the `filament_name` and
`initial_weight` already there.

Second commit is forward compatibility for the multi-tool mapping work. Both lane_data
methods guarded on `"T" in self.map`, which is a substring test that only holds while
`map` is a single string. When `map` becomes a mapping, that turns into a key-membership
test, returns False, and publishing stops with no error anywhere. `lane_index` has the
same assumption one level down (`self.map.replace(...)`), so fixing only the guard would
move the failure from a silent stop to an `AttributeError` inside the record itself.
`tool_mappings()` normalizes both forms and the guards and `lane_index` route through it.

Outer key naming is untouched here. Moving records to `T<n>` keys is your change and I
did not want to guess at its semantics. Same for `tool_mappings()` - happy to drop it in
favour of whatever #832 lands with, once that merges.

## Notes to Code Reviewers

- The string branch of `tool_mappings()` deliberately keeps the original `"T" in map`
  substring test instead of tightening it to `startswith`, so existing single-string
  configs behave exactly as they do today.
- A multi-tool lane collapses to the lowest-numbered tool for `lane_index`, sorted
  numerically so `T9` beats `T10`. There is a test pinning that, since a lexical sort
  gets it backwards.
- For a `map` with no `T` in it, `lane_index` used to return the raw string and now returns `""`. It is unreachable in production, since its only two consumers are the lane_data records and both sit behind the tool-mapping guard. Pinned with a test either way.
- `tool_mappings()` is public in case it is useful to the mapping work.
- The test fixture change is not cosmetic: `_make_afc_lane` builds lanes via
  `AFCLane.__new__`, so `spool_vendor` never existed on a test lane and the new
  publish raised `AttributeError` in 8 existing tests until the helper set it.
- Unrelated, but you may want to look at it: `./run-tests.sh --klippy` exits 0 while
  printing `X`. `--kalico` on its own correctly exits 1, so it is the combined runner
  swallowing the status. Anything gating on that exit code reads red as green.

- `ruff check .` clean
- `python -m pytest tests/` - 3316 passing
- Branch coverage (`--cov=extras --cov-branch`) on every added or changed line, with no
  partial branches in the new code
- `./run-tests.sh --klippy` - klipper passes. kalico fails `afc_htlf_commands` and
  `afc_psf_buffer_commands`, which are pre-existing: I ran the same two against a clean
  worktree at `origin/DEV` and they fail there identically.
- Not yet exercised end to end against a printer with a multi-tool lane, since the
  mapping change that produces one is not in `DEV` yet.

- [ ] Sent notification to pr-review channel requesting review (no, I talked to JimmyJon directly)

---

Disclosure per AI_USAGE_POLICY.md: drafted with Claude, then reviewed, tested and edited
by me... you know, Preston.

## Summary by CodeRabbit

* **New Features**
  * Lane data now supports lanes mapped to multiple tools and reports the lowest-numbered tool.
  * Lane information includes vendor, filament name, and initial spool weight.
  * Lane status now displays spool vendor details.
  * Clearing lane data removes associated vendor, filament, and weight information.

* **Documentation**
  * Added a changelog entry documenting updated lane metadata, spool vendor reporting, and multi-tool support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   turns the feature on (existing single-mapping behavior is unchanged until enabled).
 - Added `AFC_SWAP_MAPPING` macro to swap T(n) mappings between two lanes. Infinite spool runout now
   uses this to move a lane's mappings to the runout lane automatically.
+- `lane_data` records now carry `vendor_name`, `name` and `initial_weight`, and a lane's status reports
+  `spool_vendor`. AFC already fetched all three during a Spoolman lookup but never published them, so
+  anything reading `lane_data` could see a lane's material but not its brand. A slicer can now match a
+  lane to a brand-specific filament preset instead of falling back to the generic preset for that
+  material type (#808).
 
 ### Fixed
 - `AFC_RESET_MAPPING` now renumbers T(n) mappings sequentially instead of reusing old numbers, so
--- a/extras/AFC_lane.py
+++ b/extras/AFC_lane.py
@@ -1900,6 +1900,8 @@ class AFCLane:
                 "value": {
                     "color"          : self.color,
                     "material"       : self.material,
+                    "vendor_name"    : self.spool_vendor,
+                    "name"           : self.filament_name,
                     "bed_temp"       : self.bed_temp,
                     "nozzle_temp"    : self.extruder_temp,
                     "scan_time"      : scan_time,
@@ -1907,7 +1909,8 @@ class AFCLane:
                     "lane"           : key.replace("T", ""),
                     "extruder_index" : self.lane_extruder_index,
                     "spool_id"       : self.spool_id,
-                    "weight"         : self.weight
+                    "weight"         : self.weight,
+                    "initial_weight" : self.espooler.espooler_values.full_weight
                 }
             }
             self.afc.moonraker.send_lane_data(lane_data)
@@ -1933,6 +1936,8 @@ class AFCLane:
                 "value": {
                     "color"          : "",
                     "material"       : "",
+                    "vendor_name"    : "",
+                    "name"           : "",
                     "bed_temp"       : "",
                     "nozzle_temp"    : "",
                     "scan_time"      : "",
@@ -1940,7 +1945,8 @@ class AFCLane:
                     "lane"           : key.replace("T", ""),
                     "extruder_index" : self.lane_extruder_index,
                     "spool_id"       : None,
-                    "weight"         : 0
+                    "weight"         : 0,
+                    "initial_weight" : 0
                 }
             }
             self.afc.moonraker.send_lane_data(lane_data)
@@ -2382,6 +2388,7 @@ class AFCLane:
         response["color"]=self.color
         if not save_to_file:
             response["filament_name"] = self.filament_name
+            response["spool_vendor"] = self.spool_vendor
             response["multi_color_hexes"] = self.multi_color
             response["initial_weight"] = self.espooler.espooler_values.full_weight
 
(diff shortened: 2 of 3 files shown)


jimmyjon711:
@prestonbrown please rebase now with DEV since my PR was completed, and you should be able to get rid of the `tool_mappings` method in AFC_lane.py

prestonbrown:
Rebase is complete.

`tool_mappings()` is gone, along with the lowest-tool `lane_index` it fed - `_mapped_keys()` and the per-T(n) records cover that now. What's left is `vendor_name`/`name`/`initial_weight` in `send_lane_data` and `clear_lane_data`, plus `spool_vendor` in the live `get_status`. Squashed to one commit, and the changelog entry moved into your 2026-08-15 Added section.

One touch in your code: `_make_lane_for_get_status()` needed `lane.spool_vendor = ""` next to `filament_name`, otherwise the new `get_status` line trips `TestGetStatusMapSorting`.

Unit, klippy-klipper and klippy-kalico all green locally.

jimmyjon711 (approved):
LGTM

PR #836, opened 2026-08-19
(fix): Fix attempted moves and corrupted save state when machine is unhomed

ircmaxell:
Presently, calling `PAUSE` when the machine isn't homed (in a startup routine for example) causes the saved position state to become corrupted. Additionally, when PAUSE or RESUME are fired with an unhomed machine, they blindly attempt to move (z-hop) and cause an error which terminates the action.

## Major Changes in this PR

This PR introduces 2 changes: 

1) Only `save_pos()` if the machine is homed, and hence the current position is valid. Otherwise ignore the call.

2) On `PAUSE` and `RESUME`, check to ensure that the machine is homed before attempting any moves.

Before: 

klippy.log:
```
9:34 AM PAUSE
9:34 AM // action:paused
9:34 AM !! Must home axis first: 0.000 0.000 5.000 [0.000]
```

AFC.log
```
14:34:28 2074734.056 [AFC_error.py:cmd_AFC_PAUSE():249]         - DEBUG:AFC_PAUSE: Pausing
14:34:28 2074734.056 [AFC.py:save_pos():1113]                   - DEBUG:Saving position [0.0, 0.0, 0.0, 0.0] Base position: [0.0, 0.0, 0.0, 0.0] last_gcode_position: [0.0, 0.0, 0.0, 0.0] homing_position: [0.0, 0.0, 0.0, 0.0] speed: 25.0 speed_factor: 0.016667 absolute_coord: True absolute_extrude: True extrude_factor: 1.0
14:34:28 2074734.058 [AFC_functions.py:log_toolhead_pos():618]  - DEBUG:AFC_PAUSE: check absolute mode, POS:Position: [0.0, 0.0, 0.0, 0.0] base_position: [0.0, 0.0, 0.0, 0.0] last_position: [0.0, 0.0, 0.0, 0.0] speed: 25.0 speed_factor: 0.016667 extrude_factor: 1.0 absolute_coord: True absolute_extrude: True
```

After:

klippy.log:
```
9:37 AM PAUSE
9:37 AM // action:paused
9:37 AM // idle_timeout: Timeout set to 36000.00 s
```

AFC.log
```
14:37:41 2074926.628 [AFC_error.py:cmd_AFC_PAUSE():249]         - DEBUG:AFC_PAUSE: Pausing
14:37:41 2074926.633 [AFC_functions.py:log_toolhead_pos():618]  - DEBUG:Not Saving unhomed position, Error State: False, Is Paused False, Position_saved False, in toolchange: False, POS: Position: [0.0, 0.0, 0.0, 0.0] base_position: [0.0, 0.0, 0.0, 0.0] last_position: [0.0, 0.0, 0.0, 0.0] speed: 25.0 speed_factor: 0.016667 extrude_factor: 1.0 absolute_coord: True absolute_extrude: True
14:37:41 2074926.642 [AFC_functions.py:log_toolhead_pos():618]  - DEBUG:AFC_PAUSE: check absolute mode, POS:Position: [0.0, 0.0, 0.0, 0.0] base_position: [0.0, 0.0, 0.0, 0.0] last_position: [0.0, 0.0, 0.0, 0.0] speed: 25.0 speed_factor: 0.016667 extrude_factor: 1.0 absolute_coord: True absolute_extrude: True
14:37:41 2074926.643 [AFC_error.py:cmd_AFC_PAUSE():249]         - DEBUG:AFC_PAUSE: not moving in z cur_pos:[0.0, 0.0, 0.0, 0.0] move_z_pos:5.0
14:37:41 2074926.664 [AFC_error.py:cmd_AFC_PAUSE():249]         - DEBUG:PAUSE-Error State: False, Is Paused True, Position_saved False, in toolchange: False
```

## Summary by CodeRabbit

* **New Features**
  * Updated AFC to version 1.2.5.
  * Added support for multiple lane mappings and expanded temperature, configuration, status, and display-status information.
  * Display macros now provide updates during tool loading and unloading.
* **Bug Fixes**
  * Improved recovery messaging and tracking for failed load, unload, and tool-change operations.
  * Added warnings when lane ejection is refused.
  * Improved homing checks for safer position saving and recovery movements.
  * Prevented recovery moves when no valid saved position is available.

### DIFF
--- a/extras/AFC_error.py
+++ b/extras/AFC_error.py
@@ -170,7 +170,8 @@ class afcError:
         :param message: error message to log
         """
         self.logger.error(message)
-        if self.afc.function.is_homed(hardware_only=True) and not self.afc.function.is_paused():
+        if self.afc.function.is_homed(for_move=True) \
+            and not self.afc.function.is_paused():
             self.afc.save_pos()
             if self.pause:
                 self.pause_print()
@@ -276,7 +277,8 @@ class afcError:
         move_z_pos = self.afc.last_gcode_position[2] + self.afc.z_hop
         # Check if current position is below saved gcode position, if its lower first raise z above last saved
         #   position so that toolhead does not crash into part
-        if self.afc.function.is_homed(hardware_only=True) and self.afc.gcode_move.last_position[2] <= move_z_pos:
+        if self.afc.function.is_homed(for_move=True) \
+            and self.afc.gcode_move.last_position[2] <= move_z_pos:
             self.afc.move_z_pos(move_z_pos, "AFC_RESUME")
         else:
             self.logger.debug(f"AFC_RESUME: not moving in z cur_pos:{self.afc.gcode_move.last_position} move_z_pos:{move_z_pos}")
@@ -329,7 +331,8 @@ class afcError:
             self.afc.function.check_absolute_mode("AFC_PAUSE")
             move_z_pos = self.afc.last_gcode_position[2] + self.afc.z_hop
             # Check to see if current position is less than saved position plus z-hop
-            if self.afc.function.is_homed(hardware_only=True) and self.afc.gcode_move.last_position[2] <= move_z_pos:
+            if self.afc.function.is_homed(for_move=True) \
+                and self.afc.gcode_move.last_position[2] <= move_z_pos:
                 # Move Z up by z-hop value
                 self.afc.move_z_pos(move_z_pos, "AFC_PAUSE")
             else:
--- a/extras/AFC_error.py
+++ b/extras/AFC_error.py
@@ -276,7 +276,7 @@ class afcError:
         move_z_pos = self.afc.last_gcode_position[2] + self.afc.z_hop
         # Check if current position is below saved gcode position, if its lower first raise z above last saved
         #   position so that toolhead does not crash into part
-        if self.afc.gcode_move.last_position[2] <= move_z_pos:
+        if self.afc.function.is_homed() and self.afc.gcode_move.last_position[2] <= move_z_pos:
             self.afc.move_z_pos(move_z_pos, "AFC_RESUME")
         else:
             self.logger.debug(f"AFC_RESUME: not moving in z cur_pos:{self.afc.gcode_move.last_position} move_z_pos:{move_z_pos}")
@@ -329,7 +329,7 @@ class afcError:
             self.afc.function.check_absolute_mode("AFC_PAUSE")
             move_z_pos = self.afc.last_gcode_position[2] + self.afc.z_hop
             # Check to see if current position is less than saved position plus z-hop
-            if self.afc.gcode_move.last_position[2] <= move_z_pos:
+            if self.afc.function.is_homed() and self.afc.gcode_move.last_position[2] <= move_z_pos:
                 # Move Z up by z-hop value
                 self.afc.move_z_pos(move_z_pos, "AFC_PAUSE")
             else:
--- a/extras/AFC_error.py
+++ b/extras/AFC_error.py
@@ -292,7 +292,9 @@ class afcError:
         # The only time our resume should restore position is if there was an error that caused the pause
         if self.afc.error_state or temp_is_paused or self.afc.position_saved:
             self.set_error_state(False)
-            self.afc.restore_pos(False)
+            # Restore position if the printer is homed, otherwise leave it where it is
+            if self.afc.function.is_homed(for_move=True):
+                self.afc.restore_pos(False)
             self.pause = False
 
         self.logger.debug(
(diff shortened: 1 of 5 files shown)


jimmyjon711:
@ircmaxell please rebase against DEV branch and change PR to push into DEV instead of main

ircmaxell:
> @ircmaxell please rebase against DEV branch and change PR to push into DEV instead of main

Rebased and updated

ejsears:
@ircmaxell Can you add tests for the following functionality?
- `is_homed(hardware_only=)` Check all possible permutations
- `save_pos()` has some tests, but none of them execute with `is_homed.return_value = False`

ircmaxell:
> @ircmaxell Can you add tests for the following functionality?
> 
> * `is_homed(hardware_only=)` Check all possible permutations
> * `save_pos()` has some tests, but none of them execute with `is_homed.return_value = False`

Initial pass at tests pushed. Happy to iterate further

extras/AFC_error.py
```diff
@@ -170,7 +170,7 @@ def PauseUserIntervention(self, message: Optional[str]) -> None:
         :param message: error message to log
         """
         self.logger.error(message)
-        if self.afc.function.is_homed() and not self.afc.function.is_paused():
+        if self.afc.function.is_homed(hardware_only=True) and not self.afc.function.is_paused():
```
jimmyjon711: Can you update all your `if` statements to be like suggestion, been trying to convert all `if` statements to this format since it makes looking at diff's easier. 

Note: Do not apply my suggestion as is, because my spacing might be wrong.
```suggestion
        if (self.afc.function.is_homed(hardware_only=True)
            and not self.afc.function.is_paused()):
```

extras/AFC_functions.py
```diff
@@ -347,16 +347,18 @@ def check_homed(self, error_msg=None):
...
+    def is_homed(self, hardware_only=False):
         """
         Helper function to determine if printer is currently homed
 
+        :param hardware_only: When set to True will only return True if the printer is physically homed
         :return boolean: True if xyz is homed
         """
         curtime = self.afc.reactor.monotonic()
         kin_status = self.afc.toolhead.get_kinematics().get_status(curtime)
+        homing_check = hardware_only or not self.afc.disable_homing_check
```
jimmyjon711: This should turn into an `and` instead of a `or` so this method continues to operate the same way whenever `disable_homing_check` is set.

extras/AFC_functions.py
```diff
@@ -347,16 +347,18 @@ def check_homed(self, error_msg=None):
         else:
             return True
 
-    def is_homed(self):
+    def is_homed(self, hardware_only=False):
         """
         Helper function to determine if printer is currently homed
 
+        :param hardware_only: When set to True will only return True if the printer is physically homed
```
jimmyjon711: `hardware_only` flag seems like a weird name as you have already indicated in your discussion post in discord. Maybe renaming it to `force_check` or `require_homing` would be better. Also should put in the param comment that this variable is ignored if `disable_homing_check` is set.

jimmyjon711 (changes requested):

extras/AFC_functions.py
ircmaxell: Is that the right logic?

| hardware_only | disable_homing_check | Actually Check Machine |
| --- | --- | --- |
| True | True | True |
| True | False | True |
| False | True | False |
| False | False | True |

We want to run the check if either hardware_only is set (so we get a true state of the machine) or when homing_check is not disabled...?

extras/AFC_functions.py
ircmaxell: Force check is good, will modify. Though the behavior is the opposite, if force is True, then disable_homing_check is ignored (not the other way around), no?

extras/AFC_functions.py
jimmyjon711: Having it as an `and` should be the right logic, we want it to always return as homed (`True`) if disable homing check is set to `True` in the users config

jimmyjon711: Actually I'm probably wrong, gotta think about it for a second

extras/AFC_functions.py
jimmyjon711: This could be called pause_resume_check, since that's whats its being used for.

extras/AFC_functions.py
jimmyjon711: Ok yeah, i'm wrong since you using this to gate the save and then not do z hops in the resume.

extras/AFC_functions.py
ircmaxell: Changed to `for_move` as per Discord discussion.

extras/AFC_error.py
ircmaxell: Fixed in. the latest push

jimmyjon711 (approved):
LGTM, klippers RESUME will move the toolhead if a user homes while paused and printer was not already homed before the pause. AFC works correctly though, so maybe this is a bug in klippers code.

PR #839, opened 2026-08-22
Fix T0/CHANGE_TOOL crash after Klipper PR 7349 (#837)

### LINKED ISSUE #837: Klipper PR broke AFC getting absolute_coord variable
Update AFC to work with these klipper changes in klipper PR 7349

JPGWalterMS:
Fixes #837

Klipper PR 7349 renamed `GCodeMove.absolute_extrude` → `allow_absolute_extrude`. The `gcode_move.get_status()['absolute_extrude']` status dict key is unchanged.

AFC was still reading and writing the old instance attribute, which raises `AttributeError: 'GCodeMove' object has no attribute 'absolute_extrude'` on every T0 / CHANGE_TOOL.

This change:
- Reads via `get_status()['absolute_extrude']` when available, with getattr fallback to either attribute name
- Writes `allow_absolute_extrude` on new Klipper and `absolute_extrude` on older Klipper (`hasattr` compatibility)
- Keeps AFC's own saved copy `self.absolute_extrude` and the log/status label `absolute_extrude`

- Unit tests for `save_pos`, `restore_pos`, `log_toolhead_pos`, plus old/new Klipper helper objects

## PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool-change crashes on newer Klipper versions caused by renamed absolute-extrusion settings.
  * Improved compatibility when reading and restoring absolute-extrusion mode across current and legacy Klipper versions.
  * Added reliable fallback behavior when absolute-extrusion status is unavailable.

* **Tests**
  * Added coverage for current, legacy, and attribute-only Klipper configurations.
  * Updated position-save and restore scenarios to validate absolute-extrusion state handling.

### DIFF
--- a/extras/AFC.py
+++ b/extras/AFC.py
@@ -18,6 +18,7 @@ from typing import Dict, TYPE_CHECKING, Union, Any, Optional, Tuple, List
 if TYPE_CHECKING:
     from configfile import ConfigWrapper
     from gcode import GCodeDispatch, GCodeCommand
+    from extras.gcode_move import GCodeMove
     from extras.AFC_lane import AFCLane
     from extras.AFC_extruder import AFCExtruder
     from extras.AFC_functions import afcFunction
@@ -37,7 +38,11 @@ try: from extras.AFC_logger import AFC_logger
 except: raise error(ERROR_STR.format(import_lib="AFC_logger", trace=traceback.format_exc()))
 
 try:
-    from extras.AFC_functions import afcDeltaTime, round_floats, get_gcode_absolute_extrude, set_gcode_absolute_extrude
+    from extras.AFC_functions import (
+        afcDeltaTime, round_floats,
+        get_gcode_absolute_extrude,
+        set_gcode_absolute_extrude
+    )
 except: raise error(ERROR_STR.format(import_lib="AFC_functions", trace=traceback.format_exc()))
 
 try: from extras.AFC_utils import add_filament_switch, AFC_moonraker, AFC_PrintFileMetaData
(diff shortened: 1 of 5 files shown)


extras/AFC.py
```diff
@@ -36,7 +36,7 @@
 try: from extras.AFC_logger import AFC_logger
 except: raise error(ERROR_STR.format(import_lib="AFC_logger", trace=traceback.format_exc()))
 
-try: from extras.AFC_functions import afcDeltaTime, round_floats
+try: from extras.AFC_functions import afcDeltaTime, round_floats, get_gcode_absolute_extrude, set_gcode_absolute_extrude
```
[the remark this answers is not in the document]
JPGWalterMS: Addressed in e007b3c: try and the AFC_functions import are on separate lines (Ruff E701). Also added old-only and new-only mocks that omit get_status() while keeping the existing get_status() coverage.

extras/AFC_functions.py
```diff
@@ -64,6 +64,51 @@ def round_floats(value: Any, digits: int = 6) -> Any:
     return value
 
 
+def get_gcode_absolute_extrude(gcode_move: Any, eventtime: Optional[float] = None) -> Any:
```
jimmyjon711: Instead of typing `gcode_move` as `Any`, Please type as `GCodeMove` here and in AFC.py for `self.gcode_move`.

extras/AFC_functions.py
```diff
@@ -64,6 +64,51 @@ def round_floats(value: Any, digits: int = 6) -> Any:
...
+
+    Klipper PR 7349 renamed the internal attribute from absolute_extrude to
+    allow_absolute_extrude. The get_status() key remains 'absolute_extrude'.
+    Prefer that status key; fall back to either attribute name.
+
+    :param gcode_move: Klipper gcode_move object
+    :param eventtime: Optional event time forwarded to get_status()
+    :return Any: Current absolute-extrude flag (status key name unchanged)
+    """
+    get_status = getattr(gcode_move, "get_status", None)
```
jimmyjon711: I would rather have AFC access the variable directly instead of getting it though `get_status` since we already have the object. No need to rebuild a dictionary every time just to access variables in the gcode_move object. And changing to this would follow how its being done in the other parts of save_pos.

jimmyjon711 (changes requested):

jimmyjon711:
Forgot to add this to my review, please add a changelog entry for this fix

jimmyjon711 (approved):
Made the changes to my comments and testing on new/old klipper and kalico. Also verified that klipper crashed without these changes and now does not crash with these changes and absolure_extrude gets saved/restored properly.

PR #842, opened 2026-08-23
VIP (feat): Add virtual toolhead sensor for standalone toolheads

### LINKED ISSUE #810: [Enhancement]: For toolchangers add virtual toolhead sensor
### What are you looking to see?

Add virtual toolhead sensor for those that run toolchangers and want to run all standalone lanes but don't have a toolhead sensor.

### Why do you want this feature?

Because currently there is no way for AFC to know when filament is loaded into a toolhead when someone is using standalone lanes and does not have a toolhead sensor.

### Any other input or comments should go here.

_No response_

kiryam:
Closes #810.

## Major Changes in this PR

Adds a new `pin_tool_start: virtual` option for `[AFC_extruder]` sections. When set, AFC creates a software-only toolhead sensor for the extruder by reusing the existing `VirtualFilamentSensor` helper (the same building block the virtual bypass / FPS sensors use), instead of registering a hardware pin:

- the sensor reports filament present from startup (`tool_start_state = True`),
- no buttons/pins are registered (state can not flap or trigger false runouts),
- the sensor still shows up as `filament_switch_sensor <extruder>_tool_start` in the GUI.

This covers toolchanger setups that run standalone lanes on a toolhead without a physical toolhead sensor: the standalone lane now reports as loaded, so tool changes no longer abort with `TOOL_LOAD: Please load lane before continuing` and no manual `SET_LANE_LOADED` is needed after PREP.

## Notes to Code Reviewers

- The change is one `elif` branch in `AFCExtruder.__init__` plus the import; the `buffer` special case and hardware-pin path are untouched.
- `VirtualFilamentSensor` is instantiated with `enable_runout` mirroring the hardware path; since nothing can ever drive its state to filament-absent, the runout callback stays dormant.
- Follows the same pattern as the virtual bypass sensor, as suggested in #810.

- New unit tests in `tests/test_AFC_extruder.py` (`TestVirtualToolStart`): sensor state assumed loaded at init, no debounce button registered, and `handle_ready()` marks the standalone tc_lane as loaded.
- Full `tests/` suite passes locally (excluding the pre-existing `tests/install` characterization failures that also fail on a clean tree for me).
- Verified on real hardware: Tridex-style IDEX toolchanger, left toolhead as a standalone lane (`map: T0`) with no physical sensor. With `pin_tool_start: virtual`: klippy starts clean, PREP reports `extruder tool cmd: T0 LOADED in ToolHead`, `M104/M109 T0` resolves through the standalone lane, and T-command tool changes complete without the `Please load lane` abort.

Documentation: updated in ArmoredTurtle/AT-Documentation#188 (pin_tool_start reference in AFC_Hardware.cfg). Notification to the pr-review channel: will send once reviewers confirm the preferred channel.

---

*Note: parts of this change were written with the assistance of an AI model (GLM-5.3 by Z.ai) and were reviewed, unit-tested and verified by me on real hardware before submitting.*

## Summary by CodeRabbit

* **New Features**
  * Added support for virtual toolhead filament sensing.
  * Virtual toolheads start unloaded and disabled, with state saved and restored during preparation.
  * Filament presence can be toggled through the GUI or `SET_FILAMENT_SENSOR`.
  * Standalone load and unload operations can be disabled.
  * Sensor state stays synchronized with filament operations.
  * Virtual toolheads no longer require hardware homing.

* **Bug Fixes**
  * Improved lane state tracking, runout handling, metadata updates, and asynchronous Moonraker operations.
  * Ooze-prevention temperature checks are disabled by default for lanes sharing an extruder.

* **Documentation**
  * Updated the unreleased changelog with virtual sensor behavior.

### DIFF
--- a/extras/AFC_extruder.py
+++ b/extras/AFC_extruder.py
@@ -296,13 +296,14 @@ class AFCExtruder:
                 self.logger.info("Setting up as buffer")
             elif self.tool_start == "virtual":
                 # Virtual toolhead sensor for toolheads without a physical sensor (standalone
-                # toolchanger lanes). Filament presence is assumed and tracked in software only.
+                # toolchanger lanes). The sensor starts unloaded and disabled; the GUI switch
+                # (SET_FILAMENT_SENSOR ENABLE=) is the filament-presence state, mirroring how
+                # the virtual bypass works. The state is persisted in the vars file and restored
+                # during PREP.
                 self.fila_tool_start = VirtualFilamentSensor(
                     self.printer, f"{self.name}_tool_start", self.logger,
                     show_in_gui=self.enable_sensors_in_gui,
-                    runout_cb=self.handle_start_runout, enable_runout=self.enable_runout)
-                self.fila_tool_start.runout_helper.note_filament_present(is_filament_present=True)
-                self.tool_start_state = True
+                    enable_cb=self._virtual_tool_start_toggle)
             else:
                 self.fila_tool_start, self.debounce_button_start = add_filament_switch(f"{self.name}_tool_start", self.tool_start, self.printer,
                                                                                     self.enable_sensors_in_gui, self.handle_start_runout, self.enable_runout,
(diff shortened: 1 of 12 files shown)


extras/AFC_extruder.py
```diff
@@ -294,6 +294,15 @@ def __init__(self, config: ConfigWrapper) -> None:
...
             if self.tool_start == "buffer":
                 self.logger.info("Setting up as buffer")
+            elif self.tool_start == "virtual":
+                # Virtual toolhead sensor for toolheads without a physical sensor (standalone
+                # toolchanger lanes). Filament presence is assumed and tracked in software only.
+                self.fila_tool_start = VirtualFilamentSensor(
+                    self.printer, f"{self.name}_tool_start", self.logger,
+                    show_in_gui=self.enable_sensors_in_gui,
+                    runout_cb=self.handle_start_runout, enable_runout=self.enable_runout)
+                self.fila_tool_start.runout_helper.note_filament_present(is_filament_present=True)
```
jimmyjon711: I don't think always assuming that filament is present is a good idea. Its probably better to store this in the afc vars file and restore the state in the PREP routine when klipper restarts. This is currently how virtual bypass filament switch works.

jimmyjon711 (changes requested):

kiryam (commented):
You are right, assuming presence was wrong. Reworked in 73ba021 to mirror the virtual bypass exactly:

- the sensor starts not-loaded (sensor_enabled forced to False at creation, same as the virtual bypass switch)
- state is toggled in software via the GUI enable switch / SET_FILAMENT_SENSOR
- save_vars stores it in the vars file next to lane_loaded (system.extruders.<name>.virtual_tool_start)
- PREP restores it right where lane_loaded is restored

Verified on hardware: fresh start reports not loaded, toggling the switch flips the standalone lane to loaded, and after a klipper restart + PREP the saved state comes back.

extras/AFC_extruder.py
```diff
@@ -294,6 +294,16 @@ def __init__(self, config: ConfigWrapper) -> None:
...
+            elif self.tool_start == "virtual":
+                # Virtual toolhead sensor for toolheads without a physical sensor (standalone
+                # toolchanger lanes). The sensor starts unloaded and disabled; the GUI switch
+                # (SET_FILAMENT_SENSOR ENABLE=) is the filament-presence state, mirroring how
+                # the virtual bypass works. The state is persisted in the vars file and restored
+                # during PREP.
+                self.fila_tool_start = VirtualFilamentSensor(
+                    self.printer, f"{self.name}_tool_start", self.logger,
+                    show_in_gui=self.enable_sensors_in_gui,
+                    enable_cb=self._virtual_tool_start_toggle)
```
jimmyjon711, line 306: So a new method should not be needed to be created for this, you should be able to use the existing `note_tool_start_callback` so AFC reacts exactly the same way as a physical sensor for the virtual sensor. Which means it goes through the same loading logic to move the filament to the nozzle when the sensor is enabled.

Setting `virtual` as pin_tool_start is a big NO GO for toolheads that have filament changer attached to them, virtual should only be valid for standalone toolheads. We always need to require either a toolhead pin or needs to be set as buffer and AFC will use ramming when loading.

jimmyjon711 (changes requested):

jimmyjon711 (approved):

PR #857, opened 2026-09-01
show alternate install instructions

### LINKED ISSUE #855: [Bug]: crudini requirement on a Creality K1
### What happened?

If crudini is not found it throws a error and tells the user to use apt to install crudini, probably what would be better is to just use pip to install crudini, which is all I did on my k1, or at least suggest that for K1 users rather than mentioning apt.

Since we don't have apt this also creates confusion for users.

### Version

latest git repo

### Relevant log output

```shell

```

ejsears:
## Major Changes in this PR

Closes #855 

Tell the user to use whatever is on their system if apt isn't available. I'm leaving it generic to not have to figure out all possible permutations. 

## Summary by CodeRabbit

- **Documentation**
  - Clarified that `jq` cannot be installed with `pip` when `apt` is unavailable.
  - Added help text for the installer’s test mode, including its development-only status and limitations.
  - Documented that test mode skips Python checks and does not clone or update the repository.


jimmyjon711 (approved):
LGTM

PR #859, opened 2026-09-03
Fix issue 806 boolean

### LINKED ISSUE #806: [Bug]: unit-level remember_spool uses config.get, so an explicit False inherits as True

ejsears:
## Major Changes in this PR

Fix issue 806 with `config.getboolean` instead of `config.get`

Closes #806 

## Notes to Code Reviewers

This is pretty basic. 

Claude made some tests.

Tests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the `remember_spool` setting so explicitly configuring it as `False` now correctly disables spool remembering.
  - Ensured `remember_spool` correctly interprets both `True` and `False` configuration values.
  - Preserved the default behavior of `remember_spool` being disabled when no value is provided.

- **Tests**
  - Added coverage for explicit and default `remember_spool` settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [09-03-2026]
+### Fixed
+- Setting `remember_spool` to `False` will now actually set it to `False`.
+
 ## [08-31-2026]
 ### Fixed
 - `install-afc.sh` / `update-afc.sh` now resolve the add-on directory from the script's own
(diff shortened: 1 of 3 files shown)


jimmyjon711 (approved):
LGTM, this was a hard one.

